PR #23, opened 2017-02-14
If commenting enabled on a folder, it will become the default for all children

vangheem:
@tkimnguyen please review and merge


castle/cms/browser/discussion.py
```diff
@@ -0,0 +1,60 @@
...
+
+        This method checks five different settings in order to figure out if
+        discussion is enable on a specific content object:
+
+        1) Check if discussion is enabled globally in the plone.app.discussion
+           registry/control panel.
+
+        2) Check if the allow_discussion boolean flag on the content object is
+           set. If it is set to True or False, return the value. If it set to
+           None, try further.
```
tkimnguyen: Curious how a boolean can be "None"... I guess no value has ever been set? If someone sets it to False, is there a way for them to "unset" it again to go back to taking the setting from the container?

vangheem: Not currently. Should we change it so we have 3 options, (parent, yes,  no)?

tkimnguyen: Longer term, maybe we should do something like the Sharing tab's checkbox "Inherit from above"?

vangheem: well, you can always reset to "no value" again.

tkimnguyen: 3 options (parent, yes, no) would be perfect

vangheem: I'm going to leave it the way it is right now because any further customizing of the behavior starts getting sketchy with how plone.app.discucssion works.

Issue #9, opened 2017-01-18
Castle Toolbar Accessibility: Ensures that lists are structured correctly


rnunez80:
wcag2a
wcag131

**Target:**
#edit-bar > .castle-toolbar-container > .castle-toolbar-container-side > ul

**HTML:**
`<ul data-reactid=".0.1.0">`

**Summary:**
_Fix all of the following:_
List element has direct children that are not allowed inside `<li>` elements

_Related Nodes:_
#edit-bar > .castle-toolbar-container > .castle-toolbar-container-side > ul > span

RobZoneNet:
This is fixed.

PR #39, opened 2017-04-03
fixed ticket #9. a "return ''" was adding a extra <span> in the flow …

RobZoneNet:
…of ul-li.


vangheem:
Interesting, I wouldn't have expected null to work.

Can you add a changelog entry?

vangheem:
Remember you can add your github handle to the changelog entry!

tkimnguyen:
I added Rob's handle

PR #49, opened 2017-04-05
fixing ticket 17.. named the branch wrong.  This was due to using a u…

RobZoneNet:
…rl that put the theme name into the url for the image. So it couldn't find it. Changed to what main plone uses


vangheem:
Changelog!

PR #70, opened 2017-05-03
Prevent adding duplicate tiles during install

obct537:
If the castle gets re-ran, don't re-add slot tiles if they're already there.
Also fixes an issue with the Original scale  in image tiles.


obct537:
@tkimnguyen Fair point, originally that variable was doing other things....updated

PR #73, opened 2017-05-15
Fixed issue with Original scale in image tile

obct537:


castle/cms/tiles/image.py
```diff
@@ -47,6 +54,17 @@ def get_image(self):
             return
         return self.utils.get_object(self.data['image'][0])
 
+    def get_scale(self):
+        scale = self.data.get('scale', None)
+        if scale is None:
+            return 'large'
+
+        if scale == 'orignal':
```
tkimnguyen: is that misspelled on purpose? ('orignal' vs 'original')

tkimnguyen:
ok, closing as per discussion with Sam

PR #80, opened 2017-06-02
Navtile

RobZoneNet:
Put in some very basic css for nesting


castle/cms/tiles/navigation.py
```diff
@@ -76,23 +78,79 @@ def items(self):
...
+                'obj': child[1],
+                'sub_items': []
+            })
+
+            return tree
+
+        brain_listing = [[x.getPath(), x] for x in brains]
+        sorted_brain_listing = sorted(brain_listing, key=lambda x: x[0])
+
+        limit = self.data.get('limit') or 10
```
tkimnguyen: should "10" be defined somewhere more explicitly?

obct537:
It's basically just sorting the brains by their URL's.

castle/cms/tiles/navigation.py
obct537: Eh? The "limit" setting is a required option, so it should ideally never need to use that

obct537:
@tkimnguyen Added a little more documentation, still not sure if it's clear.....but it's a tricky algorithm

PR #81, opened 2017-06-05
Fixed historyview ID

obct537:
Forgot to set the CSS ID correctly in the previous change

### DIFF
--- a/castle/cms/browser/templates/content_history.pt
+++ b/castle/cms/browser/templates/content_history.pt
@@ -76,9 +76,9 @@
                   </td>
                 </tr>
                 <tr class="historyRecord historyComment discreet" tal:condition="rhComments">
-                  <td colspan="4">
+                  <td colspan="4" tal:define="version_id item/version_id|nothing">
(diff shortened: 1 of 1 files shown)


castle/cms/browser/templates/content_history.pt
```diff
@@ -76,9 +76,9 @@
                   </td>
                 </tr>
                 <tr class="historyRecord historyComment discreet" tal:condition="rhComments">
-                  <td colspan="4">
+                  <td colspan="4" tal:define="version_id item/version_id|nothing">
                     <p i18n:translate="" tal:content="rhComments" tal:condition="rhComments"
-                       tal:attributes="id python: item.get('version_id', None)"
+                       tal:attributes="id python: 'rendered-comments-' + str(version_id) if version_id else None"
```
tkimnguyen: If that expression resolves to None what kind of value will "id" have?

obct537: There just wont be an ID, which is the intended outcome

PR #87, opened 2017-08-15
Accessibility: added title to download link

rnunez80:
Non-distinguishable links
 2.4.4 Link Purpose (In Context)

The same link text is used for links going to different destinations. Users might not know the difference if they are not somehow explained.

If the destination page is the same, this is not an issue.

If the destination pages are not the same, make sure the links can be distinguished by their link texts or WAI-ARIA labels ('aria-labelledby' or 'aria-label') alone to make the difference clear to all users.


obct537:
This is a duplicate PR

Issue #104, opened 2017-10-17
not compiling on Ubuntu 16.04 due to older version pins


polyester:
installing on Ubuntu 16.04 fails, because of these version-pins in travis.cfg:

- cryptotography = 1.1
- cffi = 1.3.0
- idna = 2.0
- pyOpenSSL = 0.15.1

upgrading these to current versions of the eggs, namely

- cryptography = 2.1.1
- cffi = 1.11.2
- idna = 2.6
- pyOpenSSL = 17.3.0 

makes install work fine, provided the libs (libssl-dev libffi-dev) are there.

Additionally, also in requirements.txt change to cryptography==2.1.1

vangheem:
This was done.

PR #105, opened 2017-10-25
Fix version pins for Linux

sebasgo:
`cryptography == 1.1.0` doesn't compile on modern Linux distributions anymore. Update this and related packages to newest version.

I have tried to install castel.cms on openSUSE Tumbleweed, @polyester did run in the same issues (see issue #104)

Testing done: Buildout, instance started and site created. Seems to work without problems. As far as I can tell the touched packages are only for the Google Analytics integration which I am not sure how test.


obct537:
@sebasgo Seems like this should all be fine. Going to test out the GA integration first though.

PR #112, opened 2017-11-12
fix some typos, add control panel descriptions

tkimnguyen:
correct typo in Interface name; add descriptions to control panels; indentation fix in README.

Was playing with PyCharm... not bad so far :)


tkimnguyen:
PyCharm's refactoring was insanely easy to [ab]use

Issue #137, opened 2018-02-02
prevent logins from public URLs


tkimnguyen:
...if public and back end URLs have been set

PR #140, opened 2018-02-03
Only allow logins from backend URLs, if set in Security panel

tkimnguyen:
… also made tweaks to field descriptions


CorySanin:
You can bypass the check for the public URL by logging in on the form at /secure-login
Also, as long as the checks are not server-sided, they can be bypassed by simpley posting the data that the server expects. Here's an example of me doing exactly that.

capture

CorySanin:
This satisfies issue #137 btw

castle/cms/browser/login.py
```diff
@@ -187,6 +189,25 @@ def login(self):
...
 
         try:
+            if hasattr(self.context,'portal_registry'):
+                backend_urls = self.context.portal_registry['plone.backend_url']
+                only_allow_login_to_backend_urls = self.context.portal_registry['plone.only_allow_login_to_backend_urls']
+                pstate = getMultiAdapter(
+                    (self.context, self.request),
+                    name=u'plone_portal_state'
+                )
+                portal_url = pstate.portal_url()
```
tkimnguyen: This is the new standard way to get the portal_url: `from plone import api; url = api.portal.get().absolute_url()` from https://docs.plone.org/develop/plone.api/docs/portal.html#get-portal-url or this older and more common `from Products.CMFCore.utils import getToolByName; portal_url = getToolByName(presenter, 'portal_url')` from https://docs.plone.org/external/plone.app.dexterity/docs/testing/mock-testing.html and https://docs.plone.org/develop/plone/misc/context.html#gettoolbyname

castle/cms/__init__.py
```diff
@@ -1,4 +1,6 @@
+from zope.i18nmessageid import MessageFactory
 
+_ = MessageFactory('castle.cms')
```
tkimnguyen: thanks for getting us on the path of translations!

castle/cms/interfaces/controlpanel.py
```diff
@@ -153,6 +143,26 @@ class ISecuritySchema(controlpanel.ISecuritySchema):
...
+    public_url = schema.TextLine(
+        title=u'Public site URL',
+        description=u'The URL the public will use to view your site.',
+        default=None,
+        required=False)
+
+    backend_url = schema.Tuple(
+        title=u'Backend site URLs',
+        description=u'The URL(s) from which you will be editing and maintaining site content. '
+                    u'One per URL per line. Main URL first.',
```
tkimnguyen: "One URL per line" (extra "per" in there). "Please enter the main URL first"  might be clearer too.

castle/cms/shield.py
```diff
@@ -37,9 +37,9 @@ def protect(req):
         protect = True
     elif setting == SHIELD.BACKEND:
         backend_url = (registry and
-                       registry.get('plone.backend_url', SHIELD.NONE) or
-                       '')
-        if backend_url.startswith(req.SERVER_URL):
+                       registry.get('plone.backend_url', []) or
```
tkimnguyen: no need to use SHIELD.NONE?

castle/cms/shield.py
```diff
@@ -37,9 +37,9 @@ def protect(req):
...
+                       [])
+        if req.SERVER_URL in backend_url:
```
tkimnguyen: is this as precise as startswith()?

castle/cms/skins/castle/login_form.cpt
```diff
@@ -66,7 +75,7 @@
                 New user?
                 </strong>
 
-                <p i18n:translate="description_no_account">
+                <p tal:condition="bad_domain" i18n:translate="description_no_account">
```
tkimnguyen: does this tal:condition belong?

castle/cms/utils.py
```diff
@@ -485,9 +485,9 @@ def get_ip(req):
 def get_backend_url():
     registry = getUtility(IRegistry)
     backend_url = registry.get('plone.backend_url', None)
-    if not backend_url:
+    if not backend_url or len(backend_url) == 0:
```
tkimnguyen: nice catch!

castle/cms/utils.py
```diff
@@ -485,9 +485,9 @@ def get_ip(req):
...
         backend_url = api.portal.get().absolute_url()
-    return backend_url
+    return backend_url[0]
```
tkimnguyen: if line 489 happens, I think this will return the first character of the portal URL

castle/cms/__init__.py
CorySanin: 😄

castle/cms/skins/castle/login_form.cpt
CorySanin: Oops I must have put that in the wrong spot 😩

castle/cms/shield.py
CorySanin: Probably not 😜

PR #148, opened 2018-03-22
Added fragments directive

obct537:
Created zcml directive so we can much more easily include new fragments into castle add-ons

### DIFF
--- a/docs/fragments.rst
+++ b/docs/fragments.rst
@@ -1,0 +1,30 @@
+Fragments
+=========
+
+Fragment description here...
+
+Fragment Directories
+--------------------
+
+To include new fragments into a Castle site using a non-theme add-on,
+throw them in a folder, and add a declaration to your `configure.zcml`:
+
+Folder structure:
+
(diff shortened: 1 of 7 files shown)


tkimnguyen:
Can we add something to the new docs on how to use this? docs/intro.rst or maybe a new file in there?

docs/fragments.rst
```diff
@@ -0,0 +1,19 @@
+Fragments
+=========
+
+Fragment description here...
+
+Fragment Directories
+--------------------
+
+To include new fragments into a Castle site using a non-theme add-on, simply
```
tkimnguyen: OK, pita mode: it's best to avoid using words like 'simply' or 'just', and do you mean a directory in the theme?

tkimnguyen: Ah, a directory in the add-on you're developing. Would be ideal to have an example directory name, where it should be relatively inside the package, and even use a real fragment and its name and the directory name.

PR #150, opened 2018-03-27
WIP: Plone 5.1

tkimnguyen:


vangheem:
Seems to me we should leapfrog 5.1 and go directly to 5.2 at this point and spend all major change effort around also the python 3 migration.

I'm closing this. Kim, please reopen if you disagree!

tkimnguyen:
Eek maybe but isn’t that a ton more work?

vangheem:
I've done quite a bit of python 2 -> 3 migrations at this point and it's really not that hard...

tkimnguyen:
Ok

PR #155, opened 2018-04-23
Add simple display type to existing content tile

lucid-0:


tkimnguyen:
Thanks @lucid-0 - what is the simple display doing differently or new?

tkimnguyen:
looks like it displays the entire body, as opposed to just the summary

PR #159, opened 2018-06-05
add status control panel

MattJHess:


buildout.cfg
```diff
@@ -3,15 +3,25 @@ extends = travis.cfg
 parts =
     instance
     celery
+    test
```
tkimnguyen: hm, maybe you should undo/revert the change to buildout.cfg (ie. you can make your buildout.cfg changes locally but don't commit them and don't push them)

buildout.cfg
```diff
@@ -3,15 +3,25 @@ extends = travis.cfg
...
 
 [celery]
 recipe = collective.recipe.celery
 broker-transport = redis
+broker-host = redis://localhost:6379/0
```
tkimnguyen: These could be useful but should be narrowed down to only what's needed

castle/cms/browser/controlpanel/status.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class StatusView(BrowserView):
+
+    def docsplit(self):
+        retval = self.DocsplitSearch()
+        if retval[0]:
+            status = 'OK'
+            icon = '<span class="glyphicon glyphicon-ok" aria-hidden="true"></span> Docsplit: '
+            return '<span>' + icon + status + '</span>'
```
tkimnguyen: It would be better to have the display logic appear in the page template. This view should return True or False (and, in the case of False, the error string).

tkimnguyen: Same for the other view methods below

castle/cms/browser/controlpanel/status.py
```diff
@@ -0,0 +1,55 @@
...
+        if not retval[0]:
+            status = 'ERROR: ' + retval[1]
+            icon = '<span class="glyphicon glyphicon-remove" aria-hidden="true"></span> Docsplit: '
+            return '<span>' + icon + status + '</span>'
+
+    def elasticsearch(self):
+        rs = redis.Redis("localhost")
```
tkimnguyen: Are we certain that redis will always be connecting to localhost? Is there a deployment case for having redis running on a different IP address from the Castle site?

castle/cms/testing.py
```diff
@@ -7,6 +7,10 @@
 from plone.app.testing import MOCK_MAILHOST_FIXTURE
 from plone.app.testing import PLONE_FIXTURE
 from plone.app.testing import PloneSandboxLayer
+from plone.app.testing import login
```
tkimnguyen: It looks to me like you added these imports but didn't need them because you added no code that use them. If that's the case, you should undo/revert the change to this file. We want to change only what is needed.

castle/cms/tests/test_status.py
```diff
@@ -0,0 +1,27 @@
...
+        login(self.portal, TEST_USER_NAME)
+        setRoles(self.portal, TEST_USER_ID, ('Member', 'Manager'))
+
+    def test_status_control_panel_exists(self):
+        view = dir(controlpanel)
+        string = 'status'
+        if string in view:
+            self.assertTrue(True, msg='None')
+        else:
+            self.assertTrue(False, msg='cant find status control panel')
```
tkimnguyen: to have the apostrophe in "can't" inside the single-quoted string you'd use two consecutive single quotes, e.g.`can''t`

tkimnguyen: The other thing we want for every change is a short entry in HISTORY.rst that describes what you're changing or adding.

castle/cms/profiles/default/controlpanel.xml
```diff
@@ -105,4 +105,17 @@
...
  </configlet>
 
+
+   <configlet
+       title="Status"
+       action_id="status-controlpanel"
+       appId="status-controlpanel"
+       category="Products"
+       condition_expr=""
+       icon_expr=""
```
obct537: Might be worth contacting @RobZoneNet about creating an icon for this panel

castle/cms/browser/controlpanel/status.py
MattJHess: Updated my page template so the logic is in there instead of my python file. Having trouble figuring out how to get the error string into the page template. Anyone know of any docs I could look at to help?

castle/cms/profiles/default/controlpanel.xml
tkimnguyen: @RobZoneNet could you share with Matt those URLs for icon choices?

castle/cms/browser/controlpanel/status.py
```diff
@@ -5,51 +5,64 @@
...
-            icon = '<span class="glyphicon glyphicon-ok" aria-hidden="true"></span> Docsplit: '
-            return '<span>' + icon + status + '</span>'
+            return True
         if not retval[0]:
-            status = 'ERROR: ' + retval[1]
-            icon = '<span class="glyphicon glyphicon-remove" aria-hidden="true"></span> Docsplit: '
-            return '<span>' + icon + status + '</span>'
+            return False
+
+    def docsplitError(self):
```
tkimnguyen: How does this method get called?

castle/cms/browser/controlpanel/status.py
```diff
@@ -5,51 +5,64 @@
...
+        retval2 = self.DocsplitSearch()
+        if not retval2[0]:
+            return str(retval2[1])
 
     def elasticsearch(self):
         rs = redis.Redis("localhost")
+        rs2 = os.environ.get('REDIS_SERVER')
```
tkimnguyen: You should also handle the case when this environment variable hasn't been set, and if it hasn't been set then you can reasonably assume the default value to be 'localhost'

castle/cms/browser/controlpanel/status.py
```diff
@@ -5,51 +5,64 @@
...
+            return True
+        if not retval[0]:
+            return False
+
+    def celeryError(self):
+        retval2 = self.CeleryChecker()
+        if 'node' in retval2[1]:
+            return retval2[1]
+        else:
+            return 'Cannot connect to Redis'
```
tkimnguyen: I think you mean 'Celery'

castle/cms/browser/controlpanel/templates/status.pt
```diff
@@ -25,15 +25,35 @@
     </div>
 
     <div id="content-core">
-        <div tal:content="structure python:view.docsplit()"/>
+        <div tal:condition="python:view.docsplit()">
```
tkimnguyen: Rather than calling view.docsplit() here and below again in line 33 you should use a tal:define in line 27 to assign view.docsplit()'s return value to a variable and then in this line 28 and in line 33 use the variable

castle/cms/browser/controlpanel/templates/status.pt
```diff
@@ -25,15 +25,35 @@
...
     </div>
     <div id="content-core">
-    <div tal:content= "structure python: view.elasticsearch()" />
+        <div tal:condition="python:view.docsplit() == False">
+            <span class="glyphicon glyphicon-remove" aria-hidden="true"></span> Docsplit: ERROR <span tal:content = "structure python: view.docsplitError()"/>
+        </div>
     </div>
     <div id="content-core">
-        <div tal:content= "structure python: view.celery()" />
+        <div tal:condition="python:view.elasticsearch()">
```
tkimnguyen: same comment here about using tal:define to store view.elasticsearch()'s return value, and similarly below for view.celery()

castle/cms/tests/test_status.py
```diff
@@ -24,4 +24,4 @@ def test_status_control_panel_exists(self):
         if string in view:
             self.assertTrue(True, msg='None')
         else:
-            self.assertTrue(False, msg='cant find status control panel')
+            self.assertTrue(False, msg='can``t find status control panel')
```
tkimnguyen: Ah, I meant that these have to be regular quote characters, not backquotes or backticks. Does this work?

castle/cms/browser/controlpanel/templates/status.pt
```diff
@@ -0,0 +1,59 @@
...
+            <span class="glyphicon glyphicon-ok" aria-hidden="true"></span> Redis: OK
+        </div>
+    </div>
+    <div id="content-core">
+        <div tal:condition="python:view.elasticsearch() == False">
+            <span class="glyphicon glyphicon-remove" aria-hidden="true"></span> Redis: ERROR
+        </div>
+    </div>
+    <div id="content-core">
+        <div tal:condition="python:view.celery()">
```
MattJHess: when the condition is true gives a TypeError: 'bool' object has no attribute '__getitem__'. (this is just a note for myself)

tkimnguyen: Maybe paste the full error stack here

castle/cms/browser/controlpanel/status.py
MattJHess: Celery needs Redis in order to work so that message is what you get when Redis is turned off. I think I should say 'Celery cannot connect to Redis' for clarity. 
screen shot 2018-06-08 at 11 28 07 am

tkimnguyen: Ah, got it!

castle/cms/browser/controlpanel/status.py
```diff
@@ -0,0 +1,79 @@
...
+import celery
+import celery.bin.base
+import celery.bin.celery
+import celery.platforms
+
+
+class StatusView(BrowserView):
+
+    def docsplit(self):
+        retval = self.DocsplitSearch()
```
lucid-0: Since self.DocsplitSearch() is returning two values, for readability it might be better to assign the values to separate names, something like `success, msg = self.DocsplitSearch()` . And in this case you can also simply `return success` (or `return retval[0]` as it's named now) without the if statements, since we are returning True if it is True and False if it is False.

castle/cms/browser/controlpanel/templates/status.pt
```diff
@@ -0,0 +1,59 @@
...
+    <div metal:use-macro="context/global_statusmessage/macros/portal_message">
+      Portal status message
+    </div>
+
+    <div id="content-core">
+        <div tal:condition="python:view.docsplit()">
+            <span class="glyphicon glyphicon-ok" aria-hidden="true"></span> Docsplit: OK
+        </div>
+    </div>
+    <div id="content-core">
```
lucid-0: The `id` attribute of html elements should be unique within the document. This explains why this might matter https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id . You should be able to get away with just having one `<div id="content-core">` to wrap all of the <div> that contain tal:condition.

castle/cms/browser/configure.zcml
```diff
@@ -727,4 +727,12 @@
     layer="..interfaces.ICastleLayer"
     />
 
+    <browser:page
```
tkimnguyen: please fix the indentation to match the other stanzas

castle/cms/browser/controlpanel/status.py
```diff
@@ -0,0 +1,49 @@
...
+
+class StatusView(BrowserView):
+
+    def docsplit(self):
+        try:
+            subprocess.check_call('docsplit')
+            x = True, 'ok'
+        except OSError as e:
+            x = False, str(e)
+        if x[0]:
```
tkimnguyen: I think can you replace lines 18-21 with just `return x`, because whether `x[0]` is true or false, that's what you're doing...

tkimnguyen: (same below)

castle/cms/browser/controlpanel/templates/status.pt
```diff
@@ -0,0 +1,49 @@
...
+
+
+    <h1> Subprocesses </h1>
+
+    <div metal:use-macro="context/global_statusmessage/macros/portal_message">
+      Portal status message
+    </div>
+
+    <div id="content-core">
+        <div tal:condition="python:view.docsplit()[0]">
```
tkimnguyen: Rather than evaluate this Python expression twice, you could wrap both divs in one more div in which you use tal:define to assign the value of the Python expression to a variable, and then in each of these contained divs you use the variable. Same applies below.

castle/cms/interfaces/controlpanel.py
```diff
@@ -503,3 +504,6 @@ class ICrawlerConfiguration(Interface):
         description=u'User agent to use when crawling sites',
         default=u'CastleCMS Crawler 1.0'
     )
+
+
+
```
tkimnguyen: Is there a commit hook tool that recommended these whitespace changes? If not, please revert them.

castle/cms/profiles/default/controlpanel.xml
```diff
@@ -71,6 +71,13 @@
     visible="True" i18n:attributes="title">
   <permission>Manage portal</permission>
  </configlet>
+  <configlet title="Status" action_id="StatusControlPanel"
```
tkimnguyen: slight indentation mismatch here

castle/cms/profiles/default/controlpanel.xml
```diff
@@ -105,4 +112,7 @@
   <permission>Plone Site Setup: Users and Groups</permission>
  </configlet>
 
+
+
+
```
tkimnguyen: no need for these extra lines; it's good to avoid making source code changes that aren't necessary

castle/cms/testing.py
```diff
@@ -211,3 +209,7 @@ def publish(self, path, basic=None, env=None, extra=None,
                        response=response)
 
         return ResponseWrapper(response, outstream, path)
+
+
+
+
```
tkimnguyen: no need for these whitespace changes

docs/controlpanel.rst
```diff
@@ -0,0 +1,7 @@
+Control Panels
+===============
+
+Status-ControlPanel:
+---------------------
+The status control panel will check subprocesses that Castle-CMS needs. This control panel checks that docsplit is
```
tkimnguyen: - please use the correct spelling of "CastleCMS"
- break out each of these tests into point form
- add elasticsearch
- rather than say that it checks that Celery is working properly, it would be more correct to say that CastleCMS is able to connect to a Celery worker

castle/cms/browser/controlpanel/templates/status.pt
tkimnguyen: Take a look at the site crawler control panel page template, https://github.com/castlecms/castle.cms/blob/59c6c52d6f5d8bf21cca92caf0e432f4312b4ab8/castle/cms/browser/controlpanel/templates/crawler.pt for ideas on how to make this one look more like the others

PR #162, opened 2018-06-20
cs/cas-265 using Castle logo when no logo is specified

CorySanin:
If a site has no site logo set, it will default to the Castle logo instead of the Plone logo.


tkimnguyen:
(we are looking into the alternative of using the `castle` skins folder to override the logo.png with castle-logo.png)

PR #163, opened 2018-06-25
(Firefox) show what is in the label attribute in a option tag

RobZoneNet:
Firefox won't show what is in the label attribute in a option tag. Because of this map creation wouldn't show the colors to choose from. So the best course of action is to make it part of the content. Also in this commit is an option to have a background color in mosaic content to not have a margin bottom so it looks like 2 or more stack items belong together

The css will probably need compiled. 

Since there is change to mosaic.xml in the default profiles this will probably have to have a upgrade step before release due to making a current castle site see the changes.


tkimnguyen:
Also need an upgrade step. OK, @obct537 or I can add that

vangheem:
@RobZoneNet just reminder to poke someone about helping to finish this up!

PR #166, opened 2018-07-10
mhcs/cas-271 Silently fail when Twitter-related keys are not found

CorySanin:


tkimnguyen:
@CorySanin I added an entry to HISTORY.rst

PR #168, opened 2018-07-12
add Event start and end datetimes to the query listing tile

tkimnguyen:
...to the query listing tile default view; also fix typos in field descriptions


tkimnguyen:
- [ ] need to add the event start/end datetimes to the other query listing views
- [ ] test what happens if event start/end datetimes are checked in Display Fields but the objects being displayed are not events

PR #177, opened 2018-08-06
Override file_url property for images

lucid-0:


vangheem:
@lucid-0 rebase and changelog please!

PR #180, opened 2018-08-09
Ss/toolbar update to allow easier inclusion of 3rd party buttons & actions

obct537:


obct537:
@tkimnguyen finally getting around to responding to your comments above. Any particular place I should add docs for that? This is a very technical, non-user-friendly system, the current readme doesn't seem to be geared towards that.

obct537:
as for testing, testing JS is a small nightmare. We need them ROBOT tests!

tkimnguyen:
@obct537 re docs you could add a new file e.g. developer.rst ?

vangheem:
@obct537 could you take a few minutes to rebase this and add some notes to developer.rst?

PR #181, opened 2018-08-09
cs/cas-285 Image Tile Changes

CorySanin:
- Moved display mode above scale in image tile options
- Scale renamed to resolution
- Rename some of the display modes, provide brief explanations of each one
- [Fixed a bug](https://github.com/castlecms/castle.cms/commit/2397772ea9c7273541495582be5307904981a98c) that occurs when resolution is set to original and display is set to one of the cropping modes
- Image tile form shows a preview of the picture's original focal point
- Changed the focal point indicator color to contrast with both light and dark backgrounds (requires recompilation of styles)


CorySanin:
I made the requested changes on this one

castle/cms/static/patterns/focalpointselect.js
```diff
@@ -85,32 +86,34 @@ define([
 
     render: function(){
       var that = this;
-      if(that.state.hidden){
```
obct537: Is this case being handled in some other way?

castle/cms/browser/imageinfo.py
```diff
@@ -0,0 +1,14 @@
+import json
+from Products.Five.browser import BrowserView
+from castle.cms import utils
+
+class ImageInfoView(BrowserView):
+
+    def __init__(self, context, request):
```
obct537: Not very important, but for future reference, this declaration isn't necessary. You should only need to override __init__ if you're doing something non-standard

CorySanin: 👌

castle/cms/static/patterns/focalpointselect.js
CorySanin: Well the image doesn't actually get hidden anymore... I suppose maybe the state property should be renamed. Though I don't know if that could have unintentional consequences.

obct537: Makes sense....I'd recommend just looking for any references to the hidden value, and verify it's not being used anywhere unexpected. Otherways, should be good

CorySanin: I didn't see it anywhere else. Gave a more descriptive name.

PR #186, opened 2018-08-15
cs/cas-218 Hide description content type fields on view

CorySanin:
You know, like you got your content type and it's got fields with descriptions... There's no reason to show the field descriptions on the view. But it _does_ makes sense to show it when editing content. [This](https://github.com/castlecms/castle.cms/commit/97b2b4251688503c7adb84ea06a462f4eb37e452) is what I changed.


tkimnguyen:
OK, this affects only what CastleCMS considers "custom" content types, because only those will use the default add/edit forms (content types included with CastleCMS will use the Mosaic editor). However: this should not affect the Dexterity content type edit form, ie. dexterity-types/MYCUSTOMCONTENTTYPE/@@fields

Please add a changelog entry.

PR #187, opened 2018-08-16
Chunked upload on edit forms for blob fields

lucid-0:
Make chunked upload component available on edit forms with blob fields


tkimnguyen:
...after several minutes' wait (no errors in console) screen shot 2018-08-28 at 10 33 14 am

castle/cms/browser/content.py
```diff
@@ -250,6 +265,32 @@ def get_type_and_location(self, info):
                 'audio_repo_location', '/audio-repository')
         return type_, location
 
+    def update_file_content(self, info):
+            type_, location = self.get_type_and_location(info)
```
vangheem: indentation here

PR #188, opened 2018-08-21
cs/cas-347 change Add New Content Type button in Dexterity control panel

CorySanin:
The Add New Content Type button now prompts the user to either select an existing content type to clone from or create a new content type from scratch via a new intermediate view.


CorySanin:
The approach I used to direct the user to the appropriate view is a bit messy. If the only way to view the form was through a modal dialog, then it could be much cleaner. But since that isn't the only way, and I wanted to show the appropriate dialog as its own modal in that case, the form has to a bit more complicated (navigate to /dexterity-types/@@add-clone-type to see what I mean). Any suggestions on how to make this solution more elegant would be greatly appreciated.

CorySanin:
> 352 Hide/remove the old 'clone content type' button

Addressed in [last commit](https://github.com/castlecms/castle.cms/pull/188/commits/645bf062ec5f5e3c54196380d4f47305a538c2e3)

castle/cms/browser/configure.zcml
```diff
@@ -164,6 +164,23 @@
...
+  <browser:page
+      name="add-clone-type"
+      for="plone.app.dexterity.interfaces.ITypesContext"
+      template="templates/add-content-type.pt"
+      permission="plone.schemaeditor.ManageSchemata"
+      layer="..interfaces.ICastleLayer"
+      />
+
+  <!--@@dexterity-types-->
+  <browser:page
```
tkimnguyen: Is this part needed?

castle/cms/browser/templates/types-listing.pt
```diff
@@ -0,0 +1,41 @@
...
+    </span>
+  </div>
+
+  <p class="crud-description documentDescription"
+      tal:condition="view/description"
+      tal:content="structure view/description">
+  </p>
+
+  <a class="pat-plone-modal"
+      href="${context/absolute_url}/@@add-clone-type">
```
tkimnguyen: I see... well, this works

tkimnguyen:
When I click on the new Toggle Advanced Options button it goes to /@@toggle-advanced which gives me a 404 at /not-found

CorySanin:
/@@toggle-advanced doesn't 404 anymore. Now all that's left is changelog and docs.

docs/advancedmode.rst
```diff
@@ -0,0 +1,25 @@
...
+
+CastleCMS by default hides several buttons/elements that most users may not need. These
+elements can be re-enabled by toggling "Advanced Mode."
+
+
+Toggling Advanced Mode
+----------------------
+
+Currently, to toggle Advanced Mode, there's a button on the Dexterity type list page. It
+is located there for now since everything that it shows/hides is related to Dextirity,
```
tkimnguyen: - typo: "Dextirity"
- mention that advanced mode is stored in a cookie or that it is per-user, per-browser?

PR #190, opened 2018-08-22
cs/cas-351 Hide Edit Content Type XML Button

CorySanin:
By default, hide the edit XML button. It will be shown if Advanced Mode is enabled (via site settings dropdown).


tkimnguyen:
- Although my advanced cookie has value 1 I can still see the "Edit XML Field Model" button on the Document type
- the location of the "Toggle Advanced" menu item should be under the personal preferences menu, as opposed to the gear menu (I realize that maybe we should move other items currently under the gear menu!)
- the "Toggle Advanced" menu item label should change depending on the presence and value of the cookie 
- I thought we were going to do more than just hide that button, e.g. change or hide the Add New Content Type button, hide other buttons
- When you're looking at a type's fields, e.g. http://localhost:8080/Castle/dexterity-types/Document/@@fields you can't toggle the mode (advanced or not); can we add the top menu bar back into this view easily?

CorySanin:
I'll be adding docs to #188 instead. That's where the Advanced toggle is now.

PR #192, opened 2018-08-27
add copyright year to footer of new sites at create time

tkimnguyen:


tkimnguyen:
Plone does it here https://github.com/plone/Products.CMFPlone/blob/56d33f2c6b9ccb95456d54729f11c63e308de6c7/Products/CMFPlone/browser/templates/footer.pt and I think it would be better if we did it in our theme with a TAL expression, but hey... at least we have a copyright year now by default, even if it is static (the year the site was created)

PR #196, opened 2018-08-28
cs/cas-218 more precise way of hiding descriptions

CorySanin:
This will hopefully be the last that you see of this ticket in a while


tkimnguyen:
I don't think the @@ is necessary; you can go to the view without the @@ prefix.

PR #199, opened 2018-08-30
Added unicode catches for querylisting tile

obct537:


castle/cms/tiles/querylisting.py
```diff
@@ -269,7 +269,16 @@ def filter_pattern_config(self):
...
-        return json.dumps(config)
+        
+        out = '{}'
+        try:
+            out = json.dumps(config)
+        except UnicodeError:
+            try:
+                out = json.dumps(config, ensure_ascii=False)
+            except UnicodeError:
+                return '{}'
```
CorySanin: This is nitpicking, but this line could just set out to `{}` (or just `pass` maybe) and you'd only have one exit point. That's my only feedback.

PR #203, opened 2018-09-21
Updates for twitter embedding

lucid-0:


castle/cms/tiles/twittertimeline.py
```diff
@@ -11,40 +11,21 @@
 
 class TwitterTimelineTile(BaseTile):
     def render(self):
-        screenName = self.data.get('screenName', '')
+        self.screenName = self.data.get('screenName', '')
```
vangheem: `self.screenName`?

Doesn't seem to be assigning instance members on render.

If you need these in the template, you can do `view/data/screenName`

PR #208, opened 2018-09-29
cs/cas-358 fix startup permission warnings

CorySanin:
Get rid of some deprecated warnings when starting up Castle instance


vangheem:
@CorySanin can you rebase this please?

PR #211, opened 2018-10-02
Allow trailing slashes in backend URLs

CorySanin:


tkimnguyen:
I was saying to Cory it would be nice to have a way of validating those URLs before they get saved into the registry, or of removing trailing '/' before they're saved into the registry.

CorySanin:
\#wishlist

But no, it's not related to proxy rewrites. It has everything to do with user input like Kim said.

PR #217, opened 2018-10-16
Improve file edit upload

vangheem:
Less simple than I hoped because we need to hook into multiple places.

Here is a video of the implementation:
https://www.dropbox.com/s/lpwu5dkxd78ggrc/file-upload-fix-castle.mov?dl=0


vangheem:
Did an in-person review on this will @lucid-0

PR #221, opened 2018-10-23
Rp/query listing accessibility fix

RobZoneNet:
A div tag shouldn't be within a p tag. It causes issues with the DOM


vangheem:
With how scales are used though, the lower resolution allows us to scale up if necessary. It helps with performance on mobile.

PR #222, opened 2018-10-23
cs/cas-194 - Improve Management of Subscribers and Categories in Announcements

CorySanin:
Changes

* [Fixed a bug](https://github.com/castlecms/castle.cms/pull/222/commits/9b9324f875aab0597820602c2273c086eb848956#diff-6fdfa661c2a605722947a1e4461525f6R187) where it was assumed that a Python array would be passed in, when in actuality it gets a string
* [Fixed a bug](https://github.com/castlecms/castle.cms/pull/222/commits/753d758337af2d676c4911cc92541f30a729ce79#diff-1c2ded24b22c586e0aab78377cb26066R30) where iterating through subscribers could result in the loop exiting too soon (and stop sending emails)
* Added a new tab to announcement page to house all the new features
* Added an export option, which allows for exporting subscribers to an CSV, optionally by categories
* Added a list of announcement categories, as well as the number of subscribers each one has
* Added sections for adding/changing/removing categories that update subscribers' subscriptions


castle/cms/browser/controlpanel/announcements.py
```diff
@@ -220,6 +226,215 @@ def handle_text(self, action):
...
+        if not errors:
+            response = self.request.response
+            cd = 'attachment; filename=subscribers.csv'
+            response.setHeader('Content-Disposition', cd)
+            fields = ['name', 'email', 'phone_number', 'phone_number_confirmed',
+                      'confirmed', 'code', 'created', 'captcha', 'categories']
+            responsebody = ','.join(fields)
+            categories = set()
+            if 'form.widgets.export_categories' in self.request.form:
+                if self.request.form['form.widgets.export_categories'] != u'':
```
vangheem: I'm curious, aren't these values available on the `data` object? Maybe `data['export_categories']` instead of `self.request.form['form.widgets.export_categories']`. It should also deserialize the value for you too.

castle/cms/browser/controlpanel/announcements.py
```diff
@@ -220,6 +226,215 @@ def handle_text(self, action):
...
+
+    @button.buttonAndHandler(u'Rename/Merge', name='merge')
+    def handle_merge(self, action):
+        data, errors = self.extractData()
+        if not errors:
+            allcategories = api.portal.get_registry_record(reg_key)
+            categories = set()
+            newname = u''
+            if 'form.widgets.rename_merge_categories' in self.request.form:
+                if self.request.form['form.widgets.rename_merge_categories'] != u'':
```
vangheem: Same here

castle/cms/browser/controlpanel/announcements.py
```diff
@@ -220,6 +226,215 @@ def handle_text(self, action):
...
+                        'new_category_name',
+                        Invalid(
+                            u"That category name is already in use"
+                        )
+                    )
+                    return
+                for category in categories:
+                    allcategories.remove(category)
+                allcategories.append(newname)
+                api.portal.set_registry_record(reg_key, allcategories)
```
vangheem: nice.

castle/cms/browser/controlpanel/announcements.py
```diff
@@ -220,6 +226,215 @@ def handle_text(self, action):
...
+
+    ignoreContext = True
+
+    @button.buttonAndHandler(u'Delete Categories', name='delete')
+    def handle_delete(self, action):
+        data, errors = self.extractData()
+        if not errors:
+            allcategories = api.portal.get_registry_record(reg_key)
+            categories = set()
+            if 'form.widgets.delete_categories' in self.request.form:
```
vangheem: same comment as before with form data.

castle/cms/tasks/email.py
```diff
@@ -27,7 +27,7 @@ def send_email_to_subscribers(subject, html, categories=None, sender=None):
                 # make sure that this message falls under one of
                 # their categories
                 if len(categories.intersection(subscriber['categories'])) == 0:
-                    return
+                    continue
```
vangheem: ouch, nice catch.

castle/cms/browser/controlpanel/announcements.py
CorySanin: I stole it from [here](https://github.com/castlecms/castle.cms/blob/master/castle/cms/browser/controlpanel/announcements.py#L186). Looks way nicer (and makes more sense) using data. Commit incoming.

CorySanin:
All fixed!

PR #229, opened 2018-10-30
change "Castle" to "CastleCMS"

tkimnguyen:
change occurrences of "Castle" to "CastleCMS"


tkimnguyen:
Tests had passed, though now probably re-running because I added a changelog entry.

PR #237, opened 2018-11-06
fix the %20 issue that was put into the data-type.

RobZoneNet:
This caused the type to get confused and you started over essentially with "adding a type".


castle/cms/fragments/templates/dashboard-welcome.pt
```diff
@@ -21,7 +21,7 @@
    <img src="${user_url}" />
  </div>
   <p class="welcome-message">
-    What do you need to do today?<br/>
+    What do you need to do todayZ?<br/>
```
vangheem: Please remove?

PR #243, opened 2018-11-13
cs/cas-356 HIBP Integration

CorySanin:
Before updating an account's password, check if the password has been included in [HIBP](https://haveibeenpwned.com/)'s database of leaked passwords. If so, reject the password and request a new one. Setting can be enabled through the security panel.

Oh yeah, and there are tests included!


castle/cms/passwordvalidation/hibp.py
```diff
@@ -0,0 +1,49 @@
...
+        # Permit empty passwords here to allow registration links.
+        # Plone will force the user to set one.
+        if password is None:
+            return None
+
+        if api.user.is_anonymous():
+            # No check for registrations.
+            return None
+
+        hibp_enabled = api.portal.get_registry_record(
```
vangheem: Can provide `default=False`? Otherwise, this could throw an error in the period where the upgrade step hasn't been run yet.

castle/cms/passwordvalidation/hibp.py
```diff
@@ -0,0 +1,49 @@
...
+            'plone.hibp_enabled'
+        )
+
+        if not hibp_enabled:
+            # hibp has been disabled.
+            return None
+
+        sha = hashlib.sha1(password).hexdigest()
+        needle = sha[5:].upper()
+        url = 'https://api.pwnedpasswords.com/range/' + sha[:5]
```
vangheem: We're already using the requests library if you want to use that instead of urllib2. This is fine as well though.

castle/cms/profiles/2_1_0/browserlayer.xml
```diff
@@ -1,5 +1,5 @@
 <layers>
   <layer
     name="castle.cms.pwexpiry"
-    interface="castle.cms.pwexpiry.interfaces.ICollectivePWExpiryLayer" />
+    interface="castle.cms.interfaces.passwordvalidation.ICollectivePWExpiryLayer" />
```
vangheem: Can we remove the layer support?

We should only really be using ICastleLayer

castle/cms/profiles/2_2_0/registry/controlpanel.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
+<registry xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+          i18n:domain="plone">
+  <records interface="castle.cms.interfaces.ISecuritySchema"
```
vangheem: To upgrade only a specific field, you can do something like what is done with `file_upload_fields` instead. I imagine this doesn't overwrite any settings though so it's probably fine. However, it might be nice being explicit here so we know what the upgrade is targeting.

castle/cms/profiles/default/browserlayer.xml
```diff
@@ -5,5 +5,5 @@
     />
   <layer
     name="castle.cms.pwexpiry"
-    interface="castle.cms.pwexpiry.interfaces.ICollectivePWExpiryLayer" />
+    interface="castle.cms.interfaces.passwordvalidation.ICollectivePWExpiryLayer" />
```
vangheem: Reference previous comment, can we remove?

castle/cms/passwordvalidation/hibp.py
CorySanin: Oh I had no idea that was already being used. Yeah I'd rather use that.

castle/cms/profiles/2_1_0/browserlayer.xml
CorySanin: Is there a way to un-register a layer? Is it necessary to do so if it's possible?

vangheem: I think if you do `remote="True"`, it should remove it. You might need to test.

Some of those things requires you leave in the reference to the old layer class so it can load/import the definition. I'm not exactly sure if it is required when removing though.

castle/cms/profiles/2_2_0/registry/controlpanel.xml
CorySanin: I think I'm going to leave it as-is, since adding all the extra details when it's provided in `ISecuritySchema` already is redundant. I will, however, add a comment that specifies what change occurred. Is that a fair compromise?

castle/cms/passwordvalidation/hibp.py
CorySanin: Done.

castle/cms/profiles/default/browserlayer.xml
CorySanin: Removed.

castle/cms/profiles/2_2_0/registry/controlpanel.xml
vangheem: Sounds good!

castle/cms/profiles/2_2_0/browserlayer.xml
```diff
@@ -0,0 +1,6 @@
+<layers>
+  <layer
+    name="castle.cms.pwexpiry"
+    interface="castle.cms.pwexpiry.interfaces.ICollectivePWExpiryLayer"
+    remote="True" />
```
vangheem: typo, `remove`, not `remote`

CorySanin: Just following the [directions given](https://github.com/castlecms/castle.cms/pull/243#discussion_r233197514) 😂

senner:
rejected and removing

PR #244, opened 2018-11-14
Better integration of pwexpiry

CorySanin:
The idea is to put the ICustomPasswordValidator interface into a common location so it can be used more than once. The 8 character minimum password requirement was also moved to it's own thing. It can be used as an example for how to create a new password complexity requirements.


castle/cms/profiles/2_2_0/browserlayer.xml
```diff
@@ -0,0 +1,6 @@
+<layers>
+  <layer
+    name="castle.cms.pwexpiry"
+    interface="castle.cms.pwexpiry.interfaces.ICollectivePWExpiryLayer"
+    remove="True" />
```
vangheem: Thank you!

PR #246, opened 2018-11-16
fixing a tag that is blank for 508

RobZoneNet:
Since the carousel is behind aria-hidden, the whole thing is wisely skipped by screen readers. But still it is a requirement https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-refs.html see 2.4.4 and https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G91


tkimnguyen:
ah yeah, needs a HISTORY.rst line

tkimnguyen:
If this is it https://github.com/pixedelic/Camera/blob/master/scripts/camera.js it hasn't been updated in a while (5 years ago) and it is the same as what we have

tkimnguyen:
I checked already and the only non-reformatting line change is at 294 https://github.com/castlecms/castle.cms/pull/246/files#diff-33af257a09651c4590215455cdea9ab6R294

vangheem:
Alright, we still need a changelog but otherwise, I'm okay merging then.

PR #247, opened 2018-11-17
Make session timeout configurable

lucid-0:


castle/cms/browser/controlpanel/sessions.py
```diff
@@ -10,15 +10,22 @@ class SessionsView(BrowserView):
...
     def __call__(self):
+        if self.request.REQUEST_METHOD == 'POST':
+            if self.request.form.get('id'):
+                session_id = self.request.form.get('id')
+                user_id = self.request.form.get('userid')
+                user = api.user.get(user_id)
+                sm = SessionManager(self.context, self.request, user)
+                sm.session_id = session_id
+                sm.expire()
+            if self.request.form.get('timeout'):
```
vangheem: Wait, what is this for?

castle/cms/browser/controlpanel/templates/sessions.pt
```diff
@@ -25,6 +25,11 @@
     </div>
 
     <div id="content-core">
+      Session Inactivity Timeout:
```
vangheem: If we are going to add this configuration, it should be in the security control panel IMO.

castle/cms/browser/controlpanel/sessions.py
lucid-0: This is the code that actually changes the timeout attribute of the session plugin if they've submitted a value.  Seemed like why have a separate configuration registry key or something when the only reason for it would be to set the plugin's timeout to it's value.

vangheem: yah, we don't want a separate registry key but I think it should be configured in the security control panel. It can still modify that value though.

castle/cms/browser/controlpanel/templates/sessions.pt
lucid-0: Seems this entire sessions panel relates to security, I just thought this would be the place since it is about session configuration.  Open to moving it I suppose.

vangheem: The session control panel doesn't have any configuration though. It's for inspecting active user sessions, not for configuration anything.

I imagine there are cases where we want a type of user to be able to inspect user session(site administrator) but not configure session duration(manager).

But, I'm open to allowing whatever everyone wants here. I just thought it looked weird there.

tkimnguyen: I agree with Nathan that this setting belongs more in the security control panel. The Sessions panel is actually named Active Sessions when you're viewing it, which implies that it's meant to look at them (and yeah maybe end them). (And I'd propose renaming the panel so that it shows up as "Active Sessions" when you're looking at @@security-controlpanel)

castle/cms/browser/controlpanel/security.py
```diff
@@ -2,14 +2,29 @@
...
+from zope.schema import Int
+from plone import api
 
 class SecurityControlPanelForm(security.SecurityControlPanelForm):
     schema = ISecuritySchema
 
     def updateFields(self):
         super(SecurityControlPanelForm, self).updateFields()
         self.fields['restrict_logins_to_countries'].widgetFactory = SelectFieldWidget
+        timeout = api.portal.get().acl_users.session.timeout
```
vangheem: Clever!

tkimnguyen: Yeah, this is cool. Thanks for showing me new tricks too!

PR #253, opened 2018-11-26
add options for survey display

lucid-0:


castle/cms/browser/fourohfour.py
```diff
@@ -2,7 +2,6 @@
 from castle.cms import shield
 from castle.cms.files import aws
 from plone import api
-from Products.Five import BrowserView
```
vangheem: ooops, my bad.

PR #261, opened 2018-11-28
Ss/upgrade fix again

obct537:
oops

### DIFF
--- a/castle/cms/upgrades/upgrade_2_1_0.py
+++ b/castle/cms/upgrades/upgrade_2_1_0.py
@@ -17,3 +17,5 @@ def upgrade(context, logger=None):
 
     if backend_url is not None:
         api.portal.set_registry_record('plone.backend_url', (backend_url,))
+    else:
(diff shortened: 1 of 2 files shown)


castle/cms/upgrades/upgrade_2_1_0.py
```diff
@@ -9,7 +9,7 @@
 def upgrade(context, logger=None):
     # This value is being converted to a tuple, hold onto it or it will get
     # overwritten
-    backend_url = api.portal.get_registry_record('plone.backend_url') or ''
+    backend_url = api.portal.get_registry_record('plone.backend_url') or u''
```
vangheem: I think more proper thing here is `or None`.

We don't want a value of `(u'',)` either.

obct537: I'll buy that, I got hung up on the unicode bit...

vangheem: so then, I think you should set the value to `()` when None.

PR #274, opened 2018-12-06
Ss/tagfilter

obct537:


castle/cms/tiles/querylisting.py
```diff
@@ -92,6 +92,14 @@ class TaggedView(BaseTileView):
     tile_name = 'querylisting'
 
 
+class TagFilterView(BaseTileView):
+    name = 'tag-filter'
+    preview = '++plone++castle/images/previews/querylisting/tagged.png'
```
vangheem: missing file?

castle/cms/tiles/querylisting.py
```diff
@@ -268,6 +276,8 @@ def filter_pattern_config(self):
                 isinstance(config['query']['Subject'], basestring)):
             config['query']['Subject'] = [config['query']['Subject']]
 
+        config['display_type'] = self.data.get('display_type', None)
```
vangheem: clever using this for pattern config

castle/cms/tiles/querylisting.py
obct537: crap, good catch.

obct537: added a new preview image

PR #276, opened 2018-12-07
Use pat-imagewidget. Compile logged-in resources.

lucid-0:


castle/cms/static/patterns/imagewidget.js
```diff
@@ -168,7 +168,7 @@ define([
         selection = [this.props.reference];
       }
       var $re = $(that.refs.relateditem.getDOMNode());
-      var options = JSON.parse($('.pat-relateditems').attr('data-pat-relateditems'));
+      var options = JSON.parse($('.pat-imagewidget').attr('data-pat-imagewidget'));
```
vangheem: Ouch!

vangheem: I think we should just do a try catch instead. pat-imagewidget is wrong here.

PR #260, opened 2018-11-27
cs/cas-414 Import Announcement Subscribers

CorySanin:
import subscribers from a csv generated through the announcements panel. This is an extension to the functionality added by #222.

PR #277, opened 2018-12-07
cs/cas-414 Import Announcement Subscribers... again

CorySanin:
same as #260 but hopefully without weird github glitches


castle/cms/browser/controlpanel/announcements.py
```diff
@@ -32,6 +32,10 @@
 reg_key = 'castle.subscriber_categories'
 
 
+def toUnicode(txt):
```
vangheem: Maybe you want to use `from Products.CMFPlone.utils import safe_unicode` instead?

CorySanin: will do

PR #278, opened 2018-12-10
Implement tinymce link modal button

vangheem:
Feature hidden by default for now because of what it'd be encouraging users to do.

To activate, need to add:
```
{"linkTypes": ["internal", "modallink", "external", "email", "anchor"]}
```

To TinyMCE "Other settings" value.

Here is a demo of functionality(sorry about the gif quality, I'm lazy to figure out how to tweak better):

modal-demo


castle/cms/static/patterns/modal.js
```diff
@@ -0,0 +1,136 @@
...
+        });
+      }else{
+        var pt = this.props.content.portal_type;
+        var func = this['render' + pt.charAt(0).toUpperCase() + pt.slice(1)];
+        return func();
+      }
+    }
+  });
+
+  var Modal = Base.extend({
```
obct537: For my own sanity in the future, could we maybe rename this to something that isn't the same as the castle-url/components/modal.js module?

vangheem: sure, module and class names?

PR #279, opened 2018-12-12
Add crawler interval and upgrade step

lucid-0:


castle/cms/profiles/2_3_5/registry/controlpanel.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<registry xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+          i18n:domain="plone">
+  <records interface="castle.cms.interfaces.ICrawlerConfiguration"
+           prefix="castle" />
+
+  <record name="castle.crawler_interval">
```
vangheem: You don't need this part. The lines above do the trick.

castle/cms/upgrades/upgrade_2_3_5.py
```diff
@@ -0,0 +1,19 @@
...
+from zope.component import getUtility
+
+import logging
+log = logging.getLogger(__name__)
+
+PROFILE_ID = 'profile-castle.cms.upgrades:2_3_5'
+
+
+def upgrade(context, logger=None):
+    registry = getUtility(IRegistry)
```
vangheem: Instead, you just want to run the upgrade profile.

vangheem: For reference: https://github.com/castlecms/castle.cms/blob/master/castle/cms/upgrades/configure.zcml#L134

PR #280, opened 2018-12-13
ka/alias security

lucid-0:


castle/cms/shield.py
```diff
@@ -44,6 +44,19 @@ def protect(req):
                 protect |= backend_url.startswith(req.SERVER_URL)
             except AttributeError:
                 pass
-    if protect and api.user.is_anonymous():
+    if recheck:
```
vangheem: This needs to be moved below the `if protect` block so it's not done unless it's necessary.

Not every site/request will be protected so let's not do this unless we need to.

PR #291, opened 2018-12-31
Changed importer to allow add-on types

obct537:
Made the importer look for registered types, allowing add-ons to include import types.

@vangheem Simple enough, but something in the back of my head says there's a catch with this approach? Am I just senile?


obct537:
The try/except I added around the title wasn't part of this (fixes things for me locally)...I can remove it if it's problematic

PR #293, opened 2019-01-02
Reorganize browser module

vangheem:
- try to create some sub-modules so browser module is easier to sift through
- write more tests
- document


castle/cms/tests/robot/test_files.robot
```diff
@@ -0,0 +1,79 @@
...
+
+*** Test Cases ***
+
+Test add files
+   Given a logged-in site administrator
+   Go to  ${PLONE_URL}
+
+   Set Selenium Speed  .1
+
+   Wait Until Page Contains Element  css=.introjs-skipall
```
obct537: Does this work for you reliably? I've had a lot of trouble dealing with that damn popup

vangheem: sigh yes, I'm trying to fix this right now :(

PR #295, opened 2019-01-03
Upgrade improvements

vangheem:
- [x] upgrade castle through add-on interface
- [x] script to upgrade all sites on instance
- [x] integration tests between versions of castle
- [x] integration test data
- [x] fix organization of upgrade steps so plone detects upgrades properly

### DIFF
--- a/castle/cms/_scripts/create-site.py
+++ b/castle/cms/_scripts/create-site.py
@@ -1,0 +1,51 @@
+import argparse
+import json
+import os
+
+import transaction
+from AccessControl.SecurityManagement import newSecurityManager
+from castle.cms.cron.utils import setup_site
+
+
+try:
+    from castle.cms.browser.site.addsite import AddCastleSite
+except ImportError:
+    # required! This could be getting called from upgrade
+    # integration tests
+    from castle.cms.browser.addsite import AddCastleSite
+
+
+app = app  # noqa
+
+parser = argparse.ArgumentParser(
+    description='...')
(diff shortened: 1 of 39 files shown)


castle/cms/_scripts/create-site.py
```diff
@@ -0,0 +1,51 @@
...
+
+try:
+    from castle.cms.browser.site.addsite import AddCastleSite
+except ImportError:
+    # required! This could be getting called from upgrade
+    # integration tests
+    from castle.cms.browser.addsite import AddCastleSite
+
+
+app = app  # noqa
```
obct537: ?

vangheem: app is magic global variable inserted into script. This is to trick linter to not complain about future refs to global in code.

castle/cms/_scripts/create-site.py
```diff
@@ -0,0 +1,51 @@
...
+    description='...')
+parser.add_argument('--site-id', dest='site_id', default='Castle')
+parser.add_argument('--delete', dest='delete', default=False)
+args, _ = parser.parse_known_args()
+
+user = app.acl_users.getUser('admin')
+newSecurityManager(None, user.__of__(app.acl_users))
+
+if args.delete and args.site_id in app.objectIds():
+    app.manage_delObjects([args.site_id])
```
obct537: Feel like some feedback here would be useful....just in case

vangheem: done.

PR #301, opened 2019-01-15
import process fixes

tkimnguyen:
- [x] transition item only if it needs it
- [x] loop over all workflow chains (usually there is only one)
- [x] set the workflow history (do not add extraneous entries caused by the import process)

### DIFF
--- a/castle/cms/_scripts/importjson.py
+++ b/castle/cms/_scripts/importjson.py
@@ -346,38 +346,38 @@ def import_object(filepath):
         obj.contentLayout = importtype.layout
         importtype.post_creation(obj)
         if not args.skip_transitioning and data['state']:
-            try:
-                print('Transitioning %s to %s' % (obj.id, data['state']))
-                api.content.transition(obj, to_state=data['state'])
-            except Exception:
-                logger.error("Error transitioning %s to %s, maybe workflows"
-                                " don't match up" % (obj.id, data['state']))
-                # pass
-                if stop_if_exception:
-                    if pdb_if_exception:
-                        pdb.set_trace()
-                    raise
+            # transition item only if it needs it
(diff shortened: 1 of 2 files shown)


castle/cms/_scripts/importjson.py
```diff
@@ -346,38 +346,38 @@ def import_object(filepath):
...
-            if len(chain):
-                workflow_id = chain[0]
-                for h in review_history:
-                    wtool.setStatusOf(workflow_id, obj, h)
-                if len(chain) != 1:
-                    logger.warn('There should be 1 workflow for %s but'
-                                        'there are %i' % (path, len(chain)))
+            # loop over all workflow chains (usually only 1)
+            for workflow_id in wtool.getChainFor(obj):
+                obj.workflow_history[workflow_id] = review_history
```
vangheem: this is safe?

tkimnguyen: getChainFor() returns the workflow ID as a string. Normally we would not have more than 1 chain. It seems to work fine for the normal case and fits with the code that was there before that was taking the first element of the list returned by getChainFor().

vangheem: I guess we'll never have more than one workflow assigned. If there are 2 assigned however, it seems to me like it'll set workflow history for each

tkimnguyen: isn't that ok though? the import should result in content with matching metadata

tkimnguyen: I remember Elro's presentation in DC about some financial instruments he created that had multiple workflows attached

tkimnguyen: I've thought about it for OIE...

PR #321, opened 2019-02-18
Rp/querylistingheadingorder

RobZoneNet:


RobZoneNet:
The fix is to put in a stock "Listing" heading that is hidden from visual but not screen readers unless someone puts in a Title into the querylisting. This maintains the H3s that are on current sites and wont break their designs

PR #336, opened 2019-03-15
Rp/accessible secure login

RobZoneNet:
I tried logging into my site on the root and saw the errors since it couldn't get to the plone title.


vangheem:
Thanks Rob!

PR #370, opened 2019-11-21
482 fixing

alphaomega325:
All you need to do is to move the changes of news_item.html file.  And that will be it for the merge.


alphaomega325:
Just fixed the issues.

PR #371, opened 2019-11-22
WIP #231 News Item lead image 

cmher:
add javascript description for lead image that when actively used on a page it will not display in edit properties and if not actively used will display back on edit properties

note - the lead image behavior is a Plone default with any content type just that News Item has lead image on the layout


lucid-0:
We'll be updating plone.app.mosaic to include this message rather than here in castle.cms

PR #372, opened 2019-11-25
482 fixing

alphaomega325:
Fixed the whitespace issues.


alphaomega325:
Completed on the whitespace fixing.  You only need to merge the news_item.html.  So the whitespace fixing is just unnecessary busy work for me.

PR #379, opened 2019-12-18
575 dw failure in test

alphaomega325:
Optimized the a couple pieces of code, and added several more caches to minimize the amount of unnecessary computation that has already been done.


lucid-0:
Since this one was a tangent that did not work out just yet and won't be the priority any time soon, I'm going to close the pull request.  This leaves your branch in tact and can be re-opened if the time comes, thanks.

PR #380, opened 2020-01-07
565 dw replace docsplit dependency

alphaomega325:
Just need to send the convert.py to be committed.  Also the docsplitsubprocess isn't connected to anything in the rest of castle.cms but I do know that it is somewhat backwards compatible with previous versions.  Unfortunately I was not able to test out the docsplitsubprocess and focused my attentions on collective.documentviewer.  Which is where the real processing happens.


alphaomega325:
Merged with the 590 non-docsplit fix because there are improvements that I want to keep with the 590_v2 being my official fix for now.

PR #387, opened 2020-01-16
590 dw fillable pdf issue after metadata stripping update

alphaomega325:
Changed some commands and removed ghostscript due to 2 pass of qpdf and exiftools removed all unnecessary metadata and ghostscript doesn't help in that regard.  Also added in mutools to allow for optimization of pdfs.


alphaomega325:
Solved the flake8 issue.

castle/cms/commands.py
```diff
@@ -309,7 +345,28 @@ def convert_to_pdf(self, filepath, filename, output_dir):
...
+    loffice = LibreOfficeSubProcess()
 except IOError:
-    logger.exception("No docsplit installed.")
-    docsplit = None
+    logger.exception("Libreoffice not installed. castle.cms"
+                     "will not be able to convert text files to pdf.")
+    loffice = None
+
+
+class DocSplitSubProcess(BaseSubProcess):
```
lucid-0: is this still supposed to be called DocSplit subprocess? Could be moved somewhere now that we won't be using docsplit.

castle/cms/interfaces/cache.py~
```diff
@@ -0,0 +1,17 @@
+from zope import schema
```
lucid-0: I see your addition of *~ to .gitignore was that intended to exclude this file?

travis.cfg
```diff
@@ -32,6 +32,7 @@ eggs +=
     castle.cms[test]
     celery[sqlalchemy]
     collective.celery
+    collective.documentviewer
```
lucid-0: we don't want to commit this change, as documentviewer will still generally be installed as it is currently.

castle/cms/interfaces/cache.py~
alphaomega325: Yes it was suppose to ignore all backup files that emacs automatically create.

castle/cms/commands.py
alphaomega325: I did this to provide backwards compatibility with any possible addons that have been added on by our users.

alphaomega325: Or to be more accurately, users that use castle.cms but doesn't use our services.

travis.cfg
alphaomega325: Just removed it.

castle/cms/interfaces/cache.py~
alphaomega325: And removed.

PR #390, opened 2020-01-29
Castle #225   email subscription

DavidHenne:


castle/cms/subscribe.py
```diff
@@ -25,17 +25,29 @@ def __init__(self):
...
         if data is None:
             data = {}
-        data.update({
-            'created': time(),
-            'code': make_random_key(100),
-            'confirmed': False,
-            'phone_number_confirmed': False,
-            'email': email
-        })
+        if code is None:
```
lucid-0: here we can just do if code is None code = make_random_key and then remove the else clause and just do the data.update() the same either way.  It won't be more efficient cpu-wise to do it that way or anything but, little easier to read / less code.

lucid-0:
closing as duplicate https://github.com/castlecms/castle.cms/pull/393/files

PR #391, opened 2020-01-29
Video will now play at a given starting time if value is provided in …

DavidHenne:
…url.  Also appended a line to allow autoplay to function.

### DIFF
--- a/castle/cms/tiles/video.py
+++ b/castle/cms/tiles/video.py
@@ -13,6 +13,8 @@ from zope.interface import invariant
 from zope.schema.vocabulary import SimpleTerm
 from zope.schema.vocabulary import SimpleVocabulary
 
+import urlparse
+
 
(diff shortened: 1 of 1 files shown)


castle/cms/tiles/video.py
```diff
@@ -41,17 +43,24 @@ def loop(self):
     def tweak_video_url(self, url):
         # add params for properties...
         parts = url.split('/')
+        
+        # gets the 't=' start time value before cleaning the url
+        parsed = urlparse.urlparse(url)
+        start_time = urlparse.parse_qs(parsed.query)['t'][0]
```
lucid-0: If there is no t value in the url, this will likely cause a KeyError.  So we'll want to add a try/except block or an if statement to make sure 't' is there.  Otherwise we can merge this one in soon!

PR #393, opened 2020-01-31
484 subscribe tile

DavidHenne:


bduncan137:
The error messages seem good now, but it looks like you still have recaptcha off from development on line 242:

        if not self.subscriptions_enabled:

when it used to be:

        if not self.has_captcha or not self.subscriptions_enabled:

that probably needs to be put back.

PR #394, opened 2020-01-31
Replaced several metadata tags from plone to castle.cms

alphaomega325:
Just changed the plone metadata tags to castle.cms and changed the metadata.py tile so that it outputs the currently correct version of castle.cms.

### DIFF
--- a/castle/cms/tests/test_crawl.py
+++ b/castle/cms/tests/test_crawl.py
@@ -51,7 +51,7 @@ TEST_ARCHIVE_PAGE = '''
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
-<meta name="generator" content="Castle - http://www.wildcardcorp.com"/>
+<meta name="generator" content="CastleCMS - http://www.wildcardcorp.com"/>
 
 
 <!--[if IE]>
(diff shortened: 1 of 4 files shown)


castle/cms/tiles/metadata.py
```diff
@@ -85,7 +85,7 @@ def get_basic_tags(self):
                 'modificationDate': _date(context, 'modified'),
                 'publicationDate': _date(context, 'effective'),
                 'expirationDate': _date(context, 'expires'),
-                'generator': 'CastleCMS 2.0',
+                'generator': 'CastleCMS 2.5.17',
```
lucid-0: is this a way for this to be dynamic so we don't have to remember to update it each time?

castle/cms/tests/test_crawl.py
```diff
@@ -51,7 +51,7 @@ def test_crawl_page(self):
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
-<meta name="generator" content="Plone - http://plone.org"/>
+<meta name="generator" content="Castle - http://www.wildcardcorp.com"/>
```
lucid-0: I know it seems picky but we want to say CastleCMS as opposed to just Castle wherever it appears hear, we've tried to maintain that name through out the codebase

PR #395, opened 2020-02-04
Added a check for the parser to ensure that we don't have a new cal

alphaomega325:
Added a checker for the get_query function that will automatically set the defaults if multiple variables are not present or have a None value.

### DIFF
--- a/castle/cms/tiles/calendar.py
+++ b/castle/cms/tiles/calendar.py
@@ -27,6 +27,20 @@ class CalendarTile(BaseTile):
 
     def get_query(self):
         parsed = parse_query_from_data(self.data, self.context)
+        # If parsed is empty and the self.data doesn't contain any of the variables below
+        # then we have a new calendar to set the defaults on.
+        if parsed == {}:
+            if self.data.get('query', None) is None:
+                if self.data.get('limit', None) is None:
+                    if self.data.get('sort_reversed', None) is None:
+                        if self.data.get('sort_on', None) is None:
+                            self.data[u'limit'] = 250
+                            self.data[u'query'] = [{'i': 'portal_type',
+                                                    'o': 'plone.app.querystring.operation.selection.any',
+                                                    'v': ['Event']}]
+                            self.data[u'sort_reversed'] = False
+                            return self.get_query()
+
         parsed['sort_order'] = 'reverse'
         return parsed
 


alphaomega325:
Removed the hand crafted variables in the defaults and just use the defaults provided by the interface.

castle/cms/tiles/calendar.py
```diff
@@ -23,10 +24,60 @@ def format_date(dt):
...
     def get_query(self):
         parsed = parse_query_from_data(self.data, self.context)
+        # If parsed is empty and the self.data doesn't contain any of the variables below
+        # then we have a new calendar to set the defaults on.
+        if parsed == {}:
+            if self.data.get('query', None) is None:
+                if self.data.get('limit', None) is None:
+                    if self.data.get('sort_reversed', None) is None:
+                        if self.data.get('sort_on', None) is None:
+                            self.data[u'limit'] = ICalendarTileSchema["limit"].default
```
lucid-0: Here, the idea of the defaults in the schema is that they are set on the instance (the new calendar tile in this case) by default when it is created.  Such that if the user attempted to create it with a None value, the default would be used.  So I wonder if you were seeing some different behavior than that, that would cause you to do this whole block?  With the defaults that are now in the schema, this line would not likely be reached since if no value is inputted for limit for example, limit would still be 250 on line 73 rather than None, because 250 is the default set in the schema.

castle/cms/tiles/calendar.py
```diff
@@ -23,10 +24,60 @@ def format_date(dt):
...
+                            self.data[u'query'] = ICalendarTileSchema["query"].default
+                            self.data[u'sort_reversed'] = ICalendarTileSchema["sort_reversed"].default
+                            return self.get_query()
```
lucid-0: What would happen here if, for whatever reason, the value of parsed is still and empty dict { } after re-setting the defaults? That could cause some bad recursion here as each call gets to this line and adds another call to self.get_query to the stack.

castle/cms/tiles/calendar.py
```diff
@@ -23,10 +24,60 @@ def format_date(dt):
...
+
         parsed['sort_order'] = 'reverse'
```
lucid-0: From what I can see so far, (in master as well) this line here is the only one that there might be a reason to change from master, because it is setting the sort_order to 'reverse' regardless of the value of the sort_reversed field in the schema.  Since a calendar is a time-ordered thing I'm not exactly sure what the sorting here is doing but we can look into it if needed.

castle/cms/tiles/calendar.py
alphaomega325: Actually for brand new calendars the parser gets nothing and as a result will give out nothing.  The if then statements are there to check if there are a bunch of none's in the self.data.  And if there is a bunch of nones in the self.data then we know that this is a brand new calendar.  Therefore initialize the self.data with defaults.

castle/cms/tiles/calendar.py
```diff
@@ -23,10 +24,60 @@ def format_date(dt):
...
+        title=u'Limit',
+        description=u'Limit Search Results',
+        required=False,
+        default=250,
+        min=1,
+    )
+
+
 class CalendarTile(BaseTile):
+    implements(ICalendarTileSchema)
```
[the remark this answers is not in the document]
alphaomega325: I removed it and it is now working just fine, not really sure why I had it.  For the schema defaults,  I can confirm that if you remove False to any of them you will get a dialog box that will pop up informing you of your choices.  In which we really have two options, we can either have the massive if then tree followed by the defaults, or we will allow the user to be interrupted by the dialogue box.

castle/cms/tiles/calendar.py
alphaomega325: Solved it by using a for loop to ensure that the dict has no none or it will just return the first parsed.

castle/cms/tiles/calendar.py
alphaomega325: I agreed, removed it.

PR #398, opened 2020-02-10
312 dw undelete multiple objects from trash

alphaomega325:
If you want me to move the restore all button I will be willing to move it to the right.

### DIFF
--- a/castle/cms/browser/templates/trash.pt
+++ b/castle/cms/browser/templates/trash.pt
@@ -6,6 +6,8 @@
       metal:use-macro="context/main_template/macros/master"
       i18n:domain="plone">
 
+
+
 <body>
 
 <metal:b fill-slot="content-title">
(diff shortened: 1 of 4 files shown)


bduncan137:
On a quick test of deleting a page and a news item, the restore all button doesn't seem to restore either item from the recycling bin

bduncan137:
I do think, since the two button are different sizes, they would look much better side to side as opposed to on top of one another

bduncan137:
although, when i remember to develop js in development mode, it seems to work fine. Still think the buttons should be in a single line though.

bduncan137:
Also, it would be more user-friendly to label the button "restore selected items" instead of "restore all items", since that seems to imply that all items in the recycle bin, whether checked or not, would be restored.

alphaomega325:
Managed to implement your requests.  The restore all button looks funky but it allows developers to use the restore all and delete all buttons to use them below 1080 p resolution displays.  At least it should allow them until when you reach cellphone size.  I am also said to report that in order to allow the enduser to use this feature you need to compile the javascript and css code.

PR #403, opened 2020-02-21
Fixed the sticky footer issues so it only does popup animation once a…

DavidHenne:
…nd stays closed when the 'x' is clicked.

### DIFF
--- a/castle/cms/static/patterns/sticky.js
+++ b/castle/cms/static/patterns/sticky.js
@@ -19,7 +19,7 @@ define([
 
         // Sets the cookie to indicate the footer was closed and won't re-appear.
         $(".close").click(function() {
-          self.cookie_data = self.getCookie("closed-footer")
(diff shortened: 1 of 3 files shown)


castle/cms/static/patterns/sticky.js
```diff
@@ -12,13 +13,31 @@ define([
...
+                .css({ bottom: "-100px" })
+                .animate({ bottom: "0px" }, "slow");
+              self.cookie_data = self.getCookie("no-animation");
+          }
+      }
       });
-    }
+    },
+
+    getCookie: function(args) {
```
lucid-0: I think this can be renamed to setCookie since it is setting the cookie value (it must return the new value anyway such that the logic works).  Line 18 $.cookie("sticky-footer") is where it actually gets any existing value for the cookie.   I think just renaming the function and calls to it will be good.

PR #406, opened 2020-03-11
Configuration registry changes are now added to the audit log when us…

DavidHenne:
…er adds, deletes, or modifies.

### DIFF
--- a/castle/cms/subscribers.py
+++ b/castle/cms/subscribers.py
@@ -118,6 +118,10 @@ def on_object_event(obj, event):
     audit.event(obj, event)
 
 
+def on_config_modified_event(event):
+    audit.event(event)
+
+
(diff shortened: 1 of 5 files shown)


castle/cms/subscribers.py
```diff
@@ -118,6 +118,25 @@ def on_object_event(obj, event):
     audit.event(obj, event)
 
 
+def on_config_modified_event(event):
+    # Handling for theme change being triggered.
+    event_type = event.record.__name__.split('.')
+    if event_type[-2] == 'IThemeSettings':
+        on_theme_event(event, event_type)
+    else:
+        audit.event(event)
```
lucid-0: I also think for now we should wrap this in a try block with generic `except Exception:` and just log a warning that an event could not be logged.

PR #408, opened 2020-03-13
trash updates

lucid-0:


bduncan137:
Code looks good - I'm not sure how to check on the cron part of it, but it looks like it should work fine. I wonder if the button text should change based on whether the log is currently hidden or not:

` var trashButton = document.getElementById('show-trash-log');   
if(trashButton.innerHTML === 'Show Auto-Empty Log'){
          trashButton.innerHTML = 'Hide Auto-Empty Log';
        } else {
          trashButton.innerHTML = 'Show Auto-Empty Log';
        }`

I also thought the button 'Empty All Recycled Items' and 'Show Auto-Empty Log' look a bit jarring being different widths (and heights) right on top of each other. I wonder if a quick fix could be to put a <br> between the empty recycle bin form and the log button.

PR #419, opened 2020-04-10
Disallowed <script> tags to be used in HTML embed tiles.

DavidHenne:


lucid-0 (changes requested):
This one got more complicated, I think we talked about something as simple as this being alright but it turns out even `<script src="some js file">` would pass by this script tag check.  Maybe we will come back to it.

PR #421, opened 2020-04-15
fix broken links for a youtube video

bduncan137:


castle/cms/tiles/templates/video.pt
```diff
@@ -47,7 +49,8 @@
     </tal:desc>
     <p class="transcript"
         tal:condition="python: 'transcript' in df and video">
-      <a href="${video/absolute_url}/view" aria-label="Read Transcript / Download ${video/Title}">Transcript / Download</a>
+      <a tal:condition="not: youtube_url" href="${video/absolute_url}/view" aria-label="Read Transcript / Download ${video/Title}">Transcript / Download</a>
+      <a tal:condition="youtube_url" href="${video/absolute_url}/view" aria-label="Read Transcript / View ${video/Title} Source">Transcript / Visit Video Source</a>
```
lucid-0, line 53: Maybe this is where you intended to call the new utils function?  I'm not sure any new function is necessary though could probably just use the `tile_youtube_url` defined near the top of the template.

bduncan137: that would go straight to youtube, and doesn't give the user the chance to get the transcript from our video view. I'm fine with that if that's what we want to do, it's your call. 

tile_youtube_url, i'm almost certain, is an embedded link that opens up weird in a non-embed situation: https://www.youtube-nocookie.com/embed/cLNyF1Zw5tg, hence my desire for a standard youtube url.

PR #422, opened 2020-04-16
561 dw theming caching issues

alphaomega325:
Made all sorts of additions to the removal of the theme caching.  I added additions to remove caching of the themes wtithin castle.cms and the services we use outside of castle.cms.  You can see more in the various additions I have made.


alphaomega325:
Please note that I don't have access to the cloudflare server so if you want to test it I would very appreciate it if you can test cloudflare extensively and get back to me on what is happening and the log files and potential error codes for the cloudflare settings.

PR #442, opened 2020-07-01
add basic paging to archival-manage view

zombified:


castle/cms/static/components/manage-archives.js
```diff
@@ -300,12 +330,29 @@ require([
...
+      var that = this;
+      return D.div({}, [
+        D.button({
+          className: 'plone-btn plone-btn-default',
+          disabled: that.state.page <= 1,
+          onClick: that.navigatePrevClicked
+        }, 'Previous Page'),
+        D.button({
+          className: 'plone-btn plone-btn-default pull-right',
+          disabled: that.state.items.length < 100,
```
lucid-0, line 343: Should the 100 here be the per_page value which is 1000 in the python?  e.g. if there were 200 items the Next button will be enabled but no more pages available on backend?

zombified: oh yeah, thanks :) TODO: never hard code things

PR #443, opened 2020-07-06
Bd/new slideshow tweaks

bduncan137:


castle/cms/browser/content/slideshow.py
```diff
@@ -90,8 +92,19 @@ def get_resource_view_url(self, resource):
             url += '/view'
         return url
 
+    def get_domain(self):
+        portal_url = getToolByName(self.context, "portal_url")
+        portal = portal_url.getPortalObject()
+        full_url = portal.absolute_url()
```
lucid-0, line 98: just a note could use api.portal.get().absolute_url()

castle/cms/profiles/default/registry/resources.xml
```diff
@@ -223,4 +223,9 @@
     </value>
   </record>
 
+  <records prefix="plone.resources/slideshow-js"
```
lucid-0, line 226: We'll need to add this to a new upgrade step too so we don't have to re-run default profile on existing sites

castle/cms/browser/content/slideshow.py
bduncan137: that sounds better

castle/cms/profiles/default/registry/resources.xml
bduncan137: k, i'll do that now

PR #445, opened 2020-07-09
revert paging js for manage-archives, improve manage-archives perform…

zombified:
…ance with better boto3 usage

### DIFF
--- a/castle/cms/browser/controlpanel/archival.py
+++ b/castle/cms/browser/controlpanel/archival.py
@@ -173,7 +173,6 @@ class AWSApi(object):
                 log_exc=True)
 
     def list(self):
-        result = []
         base_path = archival.CONTENT_KEY_PREFIX + self.request.form.get('path', '')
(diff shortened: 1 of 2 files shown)


castle/cms/browser/controlpanel/archival.py
```diff
@@ -173,56 +173,48 @@ def save(self):
...
+                    'path': path,
+                    'id': path.rstrip('/').split('/')[-1],
+                    'is_folder': False,
+                    'url': '{endpoint_url}/{bucket}/{key}'.format(
+                        endpoint_url=self.s3.meta.client.meta.endpoint_url,
+                        bucket=self.bucket.name,
+                        key=key)
+                })
+        sorted(folderresults, key=lambda x: x['path'])
+        sorted(fileresults, key=lambda x: x['path'])
```
bduncan137, line 213: sorted() doesn't modify the iterable - probably mean `folderresults = sorted(folderresults, key=lambda x: x['path'])` and `fileresults = sorted(fileresults, key=lambda x: x['path'])`

PR #446, opened 2020-07-14
Dh/login authentication

DavidHenne:


castle/cms/browser/security/login.py
```diff
@@ -295,9 +296,29 @@ def send_authorization(self):
             'state': new_state
         })
 
+    def check_whitelist(self):
+        username = self.username
+        try:
+            white_listed = api.portal.get_registry_record('whitelist_login_domains')
+            if white_listed:
+                for wl in white_listed:
+                    if wl in username:
```
zombified, line 305: should probably check the end of the username, including the "@" separator for a whilelisted domain -- we specifically want to check for email address like usernames with domains in them here.

castle/cms/profiles/default/registry/controlpanel.xml
```diff
@@ -116,4 +116,12 @@
     </value>
   </record>
 
+  <record name="whitelist_login_domains">
+        <field type="plone.registry.field.List">
+            <title>Whitelist Login Domains</title>
+            <description>Allow users to login with email of these domains</description>
+            <value_type type="plone.registry.field.TextLine"/>
+        </field>
+  </record>
```
zombified, line 125: I think this will need an upgrade step to add to existing sites.

zombified: maybe?

castle/cms/browser/security/login.py
```diff
@@ -295,9 +296,31 @@ def send_authorization(self):
             'state': new_state
         })
 
+    def check_whitelist(self):
+        username = self.username
+        if '@' in username:
```
zombified, line 301: I suggest switching this to use [email.utils.parseaddr](https://docs.python.org/2/library/email.utils.html) instead of just splitting on "@". This will likely help take care of some of the corner cases to an email address and simplify the code a little.

castle/cms/upgrades/upgrade_2_6_18.py
```diff
@@ -0,0 +1,25 @@
...
+from plone.registry import Record
+from plone.registry.interfaces import IRegistry
+from plone import api
+from Products.CMFCore.utils import getToolByName
+from zope.component import getUtility
+
+PROFILE_ID = 'profile-castle.cms:2_6_18'
+
+
+def upgrade(context, logger=None):
```
zombified, line 11: You'll want to add a reference to the castle/cms/upgrades.zcml config, and then create an upgrade profile in castle/cms/profiles/ -- and creating the upgrade profile might mean that some of this code wouldn't be necessary.

That's the way we typically add these kinds of upgrades so far.

PR #452, opened 2020-08-17
  Dh/exclude from search

DavidHenne:


zombified (commented):
Just needs an upgrade step to make sure the 'excluded_from_search' value exists for current sites

zombified (approved):

PR #457, opened 2020-09-30
Dh/purge cache

DavidHenne:


castle/cms/caching/__init__.py
```diff
@@ -247,11 +265,25 @@ def purge(self):
                 for url in getURLsToPurge(path, settings.cachingProxies):
                     purger.purgeAsync(url)
 
-        success = True
+        success = []
         if cf.enabled:
             self.cf_enabled = True
             resp = CastlePurger.purgeSync(urls, cf)
-            success = resp.json()['success']
+            success.append({'name': 'Cloudflare', 'value': resp.json()['success']})
```
lucid-0, line 272: Maybe want a try/except here in case resp is a 4xx or 5xx and has no json() or no success key

castle/cms/profiles/2_6_18/metadata.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0"?>
+<metadata>
+  <version>2618</version>
```
lucid-0, line 3: I think the upgrade step looks good overall but we'll be on version 2.6.24 so we need to renumber/rename all the 2618 references.

lucid-0 (changes requested):

castle/cms/caching/cloudflare.py
```diff
@@ -16,37 +17,20 @@ def __init__(self):
...
             self.email is not None and
             self.zone_id is not None)
-        self.site = api.portal.get()
-        self.public_url = registry.get('plone.public_url', None)
-        if not self.public_url:
-            self.public_url = self.site.absolute_url()
-        self.site_path = '/' + self.site.virtual_url_path()
 
-    # eventual support more than just CF
     def getUrlsToPurge(self, path):
```
zombified, line 21: maybe not needed since it just calls the super method anyway?

castle/cms/caching/fastly.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self):
+        super(Fastly, self).__init__()
+        registry = getUtility(IRegistry)
+        self.fastly_token = registry.get('castle.fastly_token', None)
+        self.fastly_service_id = registry.get('castle.fastly_service_id', None)
+        self.enabled = (
+            self.fastly_token is not None and
+            self.fastly_service_id is not None)
+
+    def getUrlsToPurge(self, path):
```
zombified, line 18: Needed?

castle/cms/caching/fastly.py
```diff
@@ -0,0 +1,32 @@
...
+        return super(Fastly, self).getUrlsToPurge(path)
+
+    def purge(self, urls):
+        url = "https://api.fastly.com/service/%s/purge_all" % self.fastly_service_id
```
zombified, line 22: Ideally, we don't purge everything from the fastly cache, just the specific url's we get from getUrlsToPurge() -- does fastly have no interface to purge individual files/urls?

castle/cms/caching/stackpath.py
```diff
@@ -0,0 +1,39 @@
...
+        return super(StackPath, self).getUrlsToPurge(path)
+
+    def purge(self, urls):
+        url = "https://gateway.stackpath.com/cdn/v1/stacks/%s/purge" % self.stack_id
+        payload = {"items": [
+            {
+                "url": self.public_url,
+                "recursive": True
+            }
+        ]}
```
zombified, line 28: this seems to try and purge everything? we want just the urls from getUrlsToPurge(), ideally

zombified (changes requested):

lucid-0 (approved):
Have to update from 2.6.18 to 2.6.27 as of now, otherwise looks good although I didn't set up my own test accounts for fastly/etc.

PR #458, opened 2020-10-01
Dh/cdn url config

DavidHenne:


lucid-0 (approved):

castle/cms/interfaces/controlpanel.py
```diff
@@ -290,6 +290,39 @@ class ISiteConfiguration(Interface):
...
+        required=False,
+    )
+
+    cdn_alternate_path = schema.TextLine(
+        title=u'Configure alternate CDN path',
+        required=False,
+    )
+
+    cdn_allow_js = schema.Bool(
+        title=u'Include JS in CDN Config',
```
zombified, line 309: The title should be changed to be more direct, ie something like: `Enable CDN URL for JS resources`

This should be done for the css and image urls as well.

castle/cms/cdn.py
```diff
@@ -0,0 +1,31 @@
+from plone import api
+
+
+class cdn(object):
+
+    def __init__(self, hostname, port=80, path=''):
+        self.hostname = api.portal.get_registry_record('castle.cdn_alternate_domain') or hostname
```
zombified, line 7: hostname, port, and path here -- as configuration values, should be condensed to just one value -- a URL prefix.

The prefix should wholly replace anything in the url before the path.

IE I should be able to enter, in one field `https://cdn.example.site/prefix`

and the normal url, say `https://example.site/an/example/path`, would be transformed to `https://cdn.example.site/prefix/an/example/path`

zombified (changes requested):
I think this is a good start, especially ones the inline comments are addressed.

The next step for this PR might be to get the resource urls containing `++plone++castle` to selectively resolve, based on their type (css, js, image, and whatnot), with the CDN URL if it's configured and enabled.

Also, the site-logo should be considered an image as far as the configuration option goes -- so we should be able to serve the site logo from a cdn url. Same with the favicon.

After that, it would be good to get an update for castle.theme that would also take `++plone++castle.theme` urls into account (whether specifically, or in a more general way).

If all of those can all be done in castle.cms proper, that would be ideal.

castle/cms/interfaces/controlpanel.py
```diff
@@ -291,6 +291,32 @@ class ISiteConfiguration(Interface):
     )
 
 
+class ICDNSettings(Interface):
+    cdn_alternate_url_prefix = schema.TextLine(
```
zombified, line 295: It would be useful to make sure this value is validated as a good URL, according to what is expected by the processing method.

IE: needs scheme, needs "://" and needs a domain

PR #459, opened 2020-10-15
Dh/es custom index

DavidHenne:
WIP:


zombified (approved):

lucid-0 (changes requested):
My 2 cents was that I think we can use the same custom index field you've added here and append it to the audit log index name https://github.com/castlecms/castle.cms/blob/master/castle/cms/audit.py#L169 . We were also discussing in team channel if we might end up with a script that could avoid duplication if a custom name is added after an index already exists, we'll see.

lucid-0 (approved):

PR #460, opened 2020-10-26
reset secure flow state and retry before 403

bduncan137:
trying to fix authentication issue requiring browser refresh

### DIFF
--- a/castle/cms/browser/security/login.py
+++ b/castle/cms/browser/security/login.py
@@ -45,6 +45,7 @@ class SecureLoginView(BrowserView):
             else:
                 self.request.response.setStatus(403)
                 return json.dumps({
+                    'success': False,
                     'reason': 'Something went wrong.  Try again later.'
                 })  # this shouldn't happen, state will expire.
         else:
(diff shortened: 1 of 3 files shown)


castle/cms/browser/security/login.py
```diff
@@ -38,6 +38,10 @@ def __call__(self):
                 self.auth.set_secure_flow_state(state)
             if state in self.state_map:
                 return self.state_map[state]()
+            elif not reset:
```
lucid-0, line 41: I think the only way to get to this path in the code would be if a POST is sent from a client/browser before they've sent a GET within the last couple minutes to start a login session? Since there shouldn't be any way to get to a state that is not in the state_map. I suppose this would handle that by setting their session to the initial state and then processing their POST data as if they had been in that state.  Is that the goal?  I notice that in our example issue screenshot, we did not see the 'Something went wrong.  Try again later.' message but rather 'An unknown error occurred while attempting to login' so I wonder if it is handling that issue.

bduncan137: It seemed like (but I'm not certain) that starting to log in (POST) with an incorrect username and then on the same window without refreshing if you correct the username and POST again, there were some screwy things happening with the secure_flow_state_key such that it couldn't retrieve initial state. That is at least my best understanding. I'm willing to admit I might be mistaken about something.

'Something went wrong. Try again later.' seemed to be overwritten in secure-login.js with 'An unknown error occurred while attempting to login'.

lucid-0: Okay I think that makes sense. If 2 factor is enabled and someone mistypes their username and clicks request code, they don't get the code so come back to this tab and realize the need to fix their username and click again, yeah I think this would get them back to the proper state and process their corrected input, sure.

PR #470, opened 2020-12-02
Bd/backend and frontend robots

bduncan137:
This creates configuration for telling robots how to behave on backend urls in addition to frontend ones. I can't get travis to work on even fake branches with no real changes from master, but I think this is done.


bduncan137:
Oh, and this throws in ticket 497 as a bonus - add Castle CMS version to Version Overview in control panel

DavidHenne (approved):

PR #471, opened 2020-12-02
Bd/pdf changes

bduncan137:
rearranges some pdf functionality, allows for minimal js file parsing


bduncan137:
no longer useful with updated prince changes of the last eyar

PR #474, opened 2021-01-11
Dh/nist password

DavidHenne:


zombified (changes requested):
I'd like to see some tests added specifically that test the NISTPasswordValidator.validate method, and the UsersOverviewControlPanel.set_password method. Ideally we'd also be able to test the changes to the static/components/usergroups.js file as well.

PR #475, opened 2021-01-12
made list to blacklist IPs to prevent login

BLema:
Created a list to blacklist IPs to prevent those from logging in.


BLema:
Made it so IPs get added to blacklist for failing to login several times. Also added tests for the aforementioned changes

bduncan137:
this can be closed

PR #476, opened 2021-01-28
WIP npm updates

bduncan137:
this looks to be most of the dependabot issues. Jquery is not addressed yet, but everything else is. webpack appeared to my best investigative abilities to not be used anywhere anymore. I can add that back in if it turns out i was wrong. All tests pass, but not sure what other work may need to be done to be sure there are no problems with these upgrades.


bduncan137:
jquery now removed from package.json and package-lock.json, so should silence dependabot alerts. This doesn't mean we're not using it, just that it's included through add-on packages

PR #479, opened 2021-02-18
Dh/history audit changenote

DavidHenne:


castle/cms/profiles/2_0_16/types/Event.xml
```diff
@@ -16,7 +16,7 @@
   <element value="plone.app.dexterity.behaviors.exclfromnav.IExcludeFromNavigation"/>
   <element value="plone.app.dexterity.behaviors.id.IShortName"/>
   <element value="plone.app.relationfield.behavior.IRelatedItems"/>
-  <element value="plone.app.versioningbehavior.behaviors.IVersionable" />
+  <element value="castle.cms.behaviors.versionable.IVersionable" />
```
zombified, line 19: do we need to change this in an old upgrade profile?

castle/cms/profiles/2_5_19/types/Slideshow.xml
```diff
@@ -24,7 +24,7 @@
   <element value="plone.app.relationfield.behavior.IRelatedItems"/>
   <element value="plone.app.dexterity.behaviors.metadata.IDublinCore" />
   <element value="plone.app.blocks.layoutbehavior.ILayoutAware" />
-  <element value="plone.app.versioningbehavior.behaviors.IVersionable" />
+  <element value="castle.cms.behaviors.versionable.IVersionable" />
```
zombified, line 27: also not sure this needs to be changed in an old upgrade profile?

castle/cms/subscribers.py
```diff
@@ -209,3 +217,111 @@ def on_youtube_video_state_changed(obj, event):
             tasks.youtube_video_state_changed.delay(obj)
         except CannotGetPortalError:
             pass
+
+
+def create_version_on_save(context, event):
```
zombified, line 222: was this pulled/modified from elsewhere? if it was, it'd be good to have a reference to the original in a comment

castle/cms/subscribers.py
```diff
@@ -209,3 +217,111 @@ def on_youtube_video_state_changed(obj, event):
...
+                # set, perhaps it was imported, or versioning info was
+                # inappropriately destroyed
+                create_version = True
+
+    # create new version if needed
+    if create_version:
+        pr.save(obj=context, comment=changeNote)
+
+
+def create_initial_version_after_adding(context, event):
```
zombified, line 276: was this pulled from some where? if it was, it'd be nice to have a reference to the original in a comment

zombified (approved):

castle/cms/behaviors/configure.zcml
```diff
@@ -68,5 +68,25 @@
...
+  <plone:behavior
+    name="castle.versioning"
+    title="Versioning"
+    description="Versioning support with CMFEditions"
+    provides=".versionable.IVersionable"
+    factory=".versionable.Versionable"
+    marker=".versionable.IVersioningSupport"
+    for="plone.dexterity.interfaces.IDexterityContent"
+    />
+  <subscriber
```
bduncan137, line 80: can these two <subscriber> entries be placed in castle/cms/subscribers.zcml? Also, since it looks like they're exact copies of the methods in plone.app.versioningbehavior, and I don't see anywhere where you've removed the import of plone.app.versioningbehavior, couldn't you just either a) have the handler point to the method from plone.app.versioningbehavior (if that works that way in the zcml, or if not b)in our subscribers.py, just import the two methods from plone.app.versioningbehavior, and have the zcml point to the imported methods in subscribers? It seems like it'd save a bunch of castle code if that could be done. Thoughs?

castle/cms/behaviors/versionable.py
```diff
@@ -0,0 +1,103 @@
...
+from zope import schema
+from zope.annotation.interfaces import IAnnotations
+from zope.component import adapter
+from zope.interface import alsoProvides
+from zope.interface import implementer
+from zope.interface import Interface
+from zope.publisher.interfaces.browser import IBrowserRequest
+
+
+class IVersionable(model.Schema):
```
bduncan137, line 17: Since the changenote is the only thing that changed from plone.app.versioningbehavior, I'm wondering if you could do something like this (obviously with # noqa:F401 on imports that are imported but not used in the file, and maybe not even all of those would need to be imported, idk): 

`from plone.app.versioningbehavior.behaviors import IVersionable as IOldVersionable`
`from plone.app.versioningbehavior.behaviors import IVersioningSupport`
`from plone.app.versioningbehavior.behaviors import Versionable`
`from plone.app.versioningbehavior.utils import get_change_note`

`class IVersionable(IOldVersionable):`
`    changeNote = schema.TextLine(`
`       title=_(u'label_change_note', default=u'Change Note'),`
`        description=_(u'help_change_note',`
`                     default=u'Enter a comment that describes the changes you made.'),`
`        required=True)`
`alsoProvides(IVersionable, IFormFieldProvider)`

I'm not 100% certain it would work, but it seems on the face of it that it would. Having a bunch less code in our codebase seems desirable to me, but I'm happy to have @zombified  tell me I'm wrong or it doesn't much matter.

bduncan137 (changes requested):

bduncan137 (approved):

PR #481, opened 2021-02-24
Bd/fix dexterity log puke

bduncan137:
this fixes our issue #7 - data tiles that aren't valid were attempting to render in dexterity content types control panel


bduncan137:
added to jk/resumeimport PR

PR #489, opened 2021-04-15
Bd/parallax

DavidHenne:


bduncan137 (commented):
It looks like the less file needs to be compiled and included in this pr - I couldn't get the parallax effect to work without going to development mode

bduncan137:
also, we'll need this (somehow) to be full-width (or full-width minus the toolbar on the left if logged in)

bduncan137:
i'm not sure if i'm doing something wrong, but i still am not seeing it work correctly on large screen width: 
image

bduncan137:
granted, that's very wide (3600 px), but I think we still need it to work no matter the width

DavidHenne:
So I added an adjustable width property and had the margins and text container scale accordingly so everything should remain centered and all that.  I figured this would give more flexibility if there's a case where they maybe don't want the image to be the full width of the page, but could also scale up to a very wide page if necessary.

bduncan137:
it is better, and might be ok for now, but the thing it's still missing is it's not responsive. So the image scaling is fixed and will therefore look different on different screen sizes. There's also some screen size out there (depending on the width of the chosen image, it looks like) for which it won't be full screen. That might not matter much for a normal width photo (because at 500%, the screen would need to be enormous for the picture to not span the full width of the window), but it also means that for a tiny screen, you'd be looking at a tiny piece of the image, which likely is not intuitive to content creators. @zombified , thoughts on this?

PR #491, opened 2021-04-21
created script to email the audit log

BLema:
Created a script for sending the day's audit log to a list of recipients created on the Castle CMS settings page.

### DIFF
--- a/castle/cms/cron/_email_audit.py
+++ b/castle/cms/cron/_email_audit.py
@@ -1,0 +1,70 @@
+from castle.cms import audit
+from castle.cms.utils import ESConnectionFactoryFactory
+from AccessControl.SecurityManagement import newSecurityManager
+from plone import api
+from tendo import singleton
+from Products.CMFCore.utils import getToolByName
+from castle.cms.utils import send_email
+
+import transaction
+
+
+query = {
+    "query": {
+        "range": {
+            "date": {
+                "gte": "now-1d/d"
+            }
+        }
+    }
+}
+
+def run_query():
+    registry = getToolByName(app['Castle'], 'portal_registry')
+    # index_name = 'castle'
+    index_name = audit.get_index_name()
+    es = ESConnectionFactoryFactory(registry)()
+    
+    #import pdb; pdb.set_trace()
+    results = es.search(
+        index=index_name,
+        filter_path=['hits.hits._source'],
+        doc_type=audit.es_doc_type,
+        body=query,
+        sort='date:desc')
+    
+    return results
+
+def generate_message(log):
+    html = "<p>Here is today's Audit Log:</p><ul>"
+    for entry in log:
+        html += '<li>'
+        for key, value in entry.items():
+            html += '<div>{}: {}</div>'.format(key,value)
+            
+        html += '</li>'
+    html+= '</ul>'
+    
+    return html
+
+def run(app):
+    singleton.SingleInstance('emailaudit')
+
+    user = app.acl_users.getUser('admin')  # noqa
+    newSecurityManager(None, user.__of__(app.acl_users))  # noqa
+
+    audit_log = run_query()['hits']['hits']
+    audit_log = map(
+        lambda hit: hit['_source'], 
+        audit_log
+    )
+    html = generate_message(audit_log)
+    recipients = api.portal.get_registry_record("castle.daily_audit_log_recipients", default=[])
+    transaction.begin()
+    send_email(
+            recipients, subject="Audit Log",
+            html=html)
+    transaction.commit()
+
+if __name__ == '__main__':
+    run(app)  # noqa
--- a/castle/cms/interfaces/controlpanel.py
+++ b/castle/cms/interfaces/controlpanel.py
@@ -7,6 +7,7 @@ from Products.CMFPlone.utils import validate_json
 from zope import schema
 from zope.interface import Interface
 from zope.schema.vocabulary import SimpleVocabulary
+from plone.app.users.schema import checkEmailAddress
 
 
 def create_term(val, label):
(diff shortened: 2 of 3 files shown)


bduncan137:
@BLema , I forgot to mention that, because a registry record was added, an upgrade will be required. See upgrade 2 6 27 for the standard pattern when writing upgrade code. there are about 4 or 5 files that need to be created/editted for this

castle/cms/profiles/2_6_27 copy/registry/controlpanel.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
```
bduncan137, line 1: this file needs to be renamed

castle/cms/profiles/2_6_27 copy/metadata.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0"?>
```
bduncan137, line 1: this file needs to be renamed

castle/cms/cron/_email_audit.py
```diff
@@ -0,0 +1,70 @@
...
+            "date": {
+                "gte": "now-1d/d"
+            }
+        }
+    }
+}
+
+def run_query():
+    registry = getToolByName(app['Castle'], 'portal_registry')
+    # index_name = 'castle'
```
bduncan137, line 24: commented lines can be removed

castle/cms/cron/_email_audit.py
```diff
@@ -0,0 +1,70 @@
...
+    index_name = audit.get_index_name()
+    es = ESConnectionFactoryFactory(registry)()
+    
+    #import pdb; pdb.set_trace()
```
bduncan137, line 28: commented lines can be removed

bduncan137 (changes requested):
after requested changes are made, at minimum, flake8 linter will need to be run, and code changes it recommends need to be made. Running the command `flake8 castle.cms` from the root directory will probably be sufficient, although you may need to pip install flake8 first. There may be errors it complains about that we don't care about. you can look at our specific flake8 config here https://github.com/castlecms/castle.cms/blob/master/setup.cfg, and you can see the automated tests that are run on every commit here: https://github.com/castlecms/castle.cms/blob/master/.travis.yml. you can also run our tests locally before committing with the command `bin/test`, and there are a bunch of config options that sometimes come in handy, which you can find with `bin/test --help`. Also, for the flake8 config, there are slightly different command-line args that can mimic the config in setup.cfg. For example, `flake8 --max-line-length=110 castle.cms`

castle/cms/upgrades.zcml
```diff
@@ -464,4 +464,21 @@
     handler=".upgrades.upgrade_2_6_31"
     profile="castle.cms:default"
     />
+
+  <genericsetup:registerProfile
+    name="2_6_32"
+    title="CastleCMS upgrade to 2.6.32 profile"
+    directory="profiles/2_6_32"
+    description=""
```
bduncan137, line 472: probably a good idea to throw in a short description like `2632 adds script that does something with the audit log`

castle/cms/upgrades/upgrades_2_6_32.py
```diff
@@ -0,0 +1,10 @@
+from Products.CMFCore.utils import getToolByName
```
bduncan137, line 1: A month or so back, I refactored how we do upgrades to save some code and files. For a "standard" upgrade like this, this file is now unnecessary, but we do need a line in castle/cms/upgrades/__init__.py for 2_6_32.

bduncan137:
if you click on details above, next to continuous-integration/travis-ci/pr or continuous-integration/travis-ci/push (red x indicates something in the ci tests failed), you should be able to see what specifically went wrong and where. In this case all 4 build jobs failed, and following the links, you can look through the build jobs, which after a bunch of boilerplate logging show `ConfigurationError: ('Invalid value for', 'handler', 'ImportError: Module castle.cms.upgrades has no global upgrade_2_6_32')`. This means it was unable to find an upgrade step 2_6_32, which is due to my recent refactor mentioned in another comment on the pr.

bduncan137:
now just flake8 again

bduncan137:
I'm still seeing 3 flake8 issues:
castle.cms/castle/cms/cron/_email_audit.py:23:30: F821 undefined name 'app'
castle.cms/castle/cms/cron/_email_audit.py:65:13: E126 continuation line over-indented for hanging indent
castle.cms/castle/cms/cron/_email_audit.py:71:9: F821 undefined name 'app'

the first and third aren't actually problems, since some magic makes 'app' available as this script is run. But flake8 doesn't know that, so we can tell flake8 to ignore the issue by putting `  # noqa: F821` after each of thos lines. the second just needs to have the double indent on lines 65 & 66 removed.

castle/cms/cron/_email_audit.py
```diff
@@ -0,0 +1,71 @@
...
+            "date": {
+                "gte": "now-1d/d"
+            }
+        }
+    }
+}
+
+
+def run_query():
+    registry = getToolByName(app['Castle'], 'portal_registry') # noqa: F821
```
bduncan137, line 23: in order to not create a different flake8 error, inline comments need to have exactly 2 spaces before comment

castle/cms/cron/_email_audit.py
```diff
@@ -0,0 +1,71 @@
...
+        for key, value in entry.items():
+            html += '<div>{}: {}</div>'.format(key, value)
+
+        html += '</li>'
+    html += '</ul>'
+
+    return html
+
+
+def run(app):   # noqa: F821
```
bduncan137, line 50: this wasn't where the error was - a parameter will always be defined, even if the value is None

castle/cms/cron/_email_audit.py
```diff
@@ -0,0 +1,71 @@
...
+    recipients = api.portal.get_registry_record("castle.daily_audit_log_recipients", default=[])
+    transaction.begin()
+    send_email(
+        recipients, subject="Audit Log",
+        html=html)
+    transaction.commit()
+
+
+if __name__ == '__main__':
+    run(app)
```
bduncan137, line 71: this was where app was undefined

PR #495, opened 2021-04-28
update to ES7 compatibility

zombified:


zombified:
last check not successful due to connection problem with pypi -- but the push succeeded fine, and should run the same set of tests, and that one passed.

PR #497, opened 2021-05-06
Bl/export slideshow

BLema:
Removed element value="Products.CMFPlone.interfaces.constrains.ISelectableConstrainTypes" from Slideshow.xml behaviors to solve an error preventing the exporting of slideshows. Also changed the object name to "Slideshow".


bduncan137:
i think we can close this

PR #499, opened 2021-05-17
Bl/tiles

BLema:
Created try-catch in patterns to catch an error that may happen with an unidentified tile, report it, then resume through remaining tiles and eventually continue to the page. Also created test_missing_tiles_not_available_in_settings in test_settings.py.


bduncan137:
we don't need this

PR #501, opened 2021-05-24
Bl/contenttypes

BLema:
Added new icons for all content types, removed references to content types not in castle, and added descriptions to content types that were missing one.


bduncan137:
not enough here worth keeping

PR #503, opened 2021-06-18
Bd/parallax

DavidHenne:


castle/cms/upgrades/upgrade_2_6_31.py
```diff
@@ -0,0 +1,10 @@
+from Products.CMFCore.utils import getToolByName
```
bduncan137, line 1: I don't think this file is necessary anymore?

castle/cms/browser/content/parallax.py
```diff
@@ -0,0 +1,74 @@
...
+                parallax_tile = {}
+
+                # Parallax tile properties
+                try:
+                    image = uuidToObject(data['image:list'][0])
+                    parallax_tile['image'] = image.absolute_url()
+                except Exception:
+                    parallax_tile['image'] = None
+
+                parallax_tile['bg_color'] = data.get('bg_color', ['white'])[0]
```
bduncan137, line 27: no action required here, but in the next iteration, we might allow users to type a hex code or css-recognized color (white, black, palegoldenrod, etc) in a textline with a css color validator constraint function to validate the field, and a default of 'black' or something, and similar with the text color. That should be a fairly straightforward mod if users seem interested in that.

castle/cms/upgrades/upgrade_2_6_31.py
bduncan137: it might have been old from me, idk. Let me know if i'm overlooking something, though

bduncan137 (changes requested):
change requested if there's not a need for 2_6_31.py, as i'm assuming there's not. Otherwise, g2g

bduncan137 (approved):

PR #504, opened 2021-06-25
Dh/slideshow preview

DavidHenne:


zombified (changes requested):
This is a good start, but what they really want is the ability to see a representation of what the slide will look like when it's full screen and being viewed

right now, what this change does is bring in the slide content, but doesn't actually show the slide as it would be in full screen mode.

Screen Shot 2021-07-09 at 2 35 01 PM

In the screenshot there, for example, the background images are large and the text/background relationship can't really be seen very well.

Perhaps we can specify a height/width and scale a preview to that size? Or go down a route of actually generating a rendered image that screencaps the slide as it would be viewed?

PR #525, opened 2022-05-18
style, js, contenttype, import scripts, email templates

zombified:


bduncan137:
- Update control panel icons, providing one for every castle configlet
- Update dexterity types so every one has an icon and a short desription
- Add EmailTemplate dexterity type for repeated emails
- Fix data-tiles that don't exist from rendering and spamming output
  (for dexterity type definitions in control panel in /@@dexterity-types/<DexterityTypeHere>)
- Allow resuming importjson script

bduncan137 (approved):
I approve this VERY MODIFIED from original PR

castle/cms/theming.py
```diff
@@ -54,7 +55,33 @@
 styleslot_xpath = etree.XPath('//*[@id="style_slot"]')
 dynamic_grid_xpath = etree.XPath('//*[@dynamic-grid]')
 LAYOUT_NAME = re.compile(r'[a-zA-Z_\-]+/[a-zA-Z_\-]+')
-
+meaningless_dexterity_control_panel_urls = (
```
zombified, line 58: why 'meaningless' ?

castle/cms/upgrades/__init__.py
```diff
@@ -45,7 +45,11 @@ def upgrade(context, logger=None):
 upgrade_3001 = default_upgrade_factory('3001')
 upgrade_3003 = default_upgrade_factory('3003')
 upgrade_3004 = default_upgrade_factory('3004')
-upgrade_3005 = default_upgrade_factory('3005')
+
+
+def upgrade_3005(site, logger=None):
+    # add in princexml username and password
+    re_register_profile(IAPISettings, 'castle')
```
zombified, line 52: why the change here for 3005 ?

castle/cms/theming.py
bduncan137: maybe nonexistant would be a better name? these tiles don't exist (or something they rely on doesn't exist, i'm not 100% sure) for the context that /@@dexterity-types uses, so for that context only, afaict, there is log puke saying those tiles don't render or something, to the tune of several hundred lines every time you hit that endpoint

castle/cms/profiles/3009/types/Parallax.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0"?>
+<object name="Parallax" meta_type="Dexterity FTI" i18n:domain="plone"
+   xmlns:i18n="http://xml.zope.org/namespaces/i18n">
+ <property name="title" i18n:translate="">Parallax</property>
+ <property name="description" i18n:translate="">A background that moves at a differnt rate than its text</property>
```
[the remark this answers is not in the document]
bduncan137: good catch. yeah, i'll fix that obviously

castle/cms/upgrades/__init__.py
bduncan137: it looks like it was originally a mistake to me, and i was retroactively attempting to correct it (if, for no other reason, it is not used as an incorrect template by another developer down the line). there is no profile for 3005, and the only thing the upgrade step is supposed to do is add those registry values, so that requires a re-register of an interface (crap, i named that function wrong - should be re_register_interface, not re_register_profile), not a standard upgrade function

castle/cms/theming.py
zombified: please also put a brief comment about why they are 'meaningless' or 'nonexistent' as well -- might help whomever has to work with this code in the future

castle/cms/upgrades/__init__.py
zombified: alrighty, sounds good

castle/cms/theming.py
bduncan137: k, will do

PR #526, opened 2022-06-15
add gelf log handler and script for generating user-role-permission reporting

zombified:


castle/cms/_scripts/report-users.py
```diff
@@ -0,0 +1,132 @@
...
+    extras = {}
+
+    extras["appname"] = APP_NAME
+    extras["site"] = site.id
+    extras["logid"] = LOG_ID
+    extras["reportid"] = uuid.uuid4()
+
+    users = api.user.get_users()
+    for user in users:
+        extras["username"] = user.getUserName()
```
bduncan137, line 70: this will just overwrite each previous username, correct? Is that the desired behavior?

castle/cms/_scripts/report-users.py
```diff
@@ -0,0 +1,132 @@
...
+    return args
+
+
+def report_on_users(site):
+    extras = {}
+
+    extras["appname"] = APP_NAME
+    extras["site"] = site.id
+    extras["logid"] = LOG_ID
+    extras["reportid"] = uuid.uuid4()
```
bduncan137, line 66: could just be:
```
    extras = {
        'appname':  APP_NAME,
        'site': site.id,
        'logid': LOG_ID,
        'reportid': uuid.uuid4(),
    }
```

castle/cms/_scripts/report-users.py
bduncan137: After reading through the code, it seems like the desire is just to log this data once and then not persist it in any way, so this seems like it should be fine - i'll just leave this comment in case...whatever

bduncan137 (approved):
there are a couple comments, which can probably be ignored after a full read-through. No obvious problems to me.

castle/cms/_scripts/report-users.py
zombified: sure

daxxog (approved):
LGTM

PR #529, opened 2022-07-01
WIP: Dh/google analytics

DavidHenne:


castle/cms/services/google/analytics.py
```diff
@@ -9,20 +9,27 @@ def get_ga_service(ga_scope=['https://www.googleapis.com/auth/analytics.readonly
     api_email = registry.get('castle.google_api_email', None)
     api_key = registry.get('castle.google_api_service_key_file', None)
     ga_id = registry.get('castle.google_analytics_id', None)
+    ua_id = registry.get('castle.universal_analytics_id', None)
 
-    if not api_key or not api_email or not ga_id:
+    if not api_key or not api_email or not ga_id and not ua_id:
```
bduncan137, line 14: while this is correct, i would suggest making it a bit clearer which operators should be evaluated first: `if not api_key or not api_email or (not ga_id and not ua_id):` I think it's easier to read that way. But, take or leave that suggestion

bduncan137 (changes requested):
There is one other thing i'm unclear on, and I fully admit it's likely I'm just missing something. That being said, I'm not seeing where the global site tile gets added to pages of a site. Is the idea that it's added manually to every page that wants it? Or is it added to every site page but i'm not seeing how? I would have thought (could be very wrong) that it would get added to every page in a site if the analytics info existed in the control panel. So, just looking for clarification on that, for my own knowledge if no other reason.

PR #532, opened 2022-08-16
make sure searchabletext for additionalsearchabletext implementation is entirely generated with unicode strings

zombified:


castle/cms/indexing/search.py
```diff
@@ -48,17 +54,25 @@ def __call__(self, hpscatalog, existing_data):
                 data[key + '_shares'] = value
         sdata = ISearch(self.obj, None)
         if sdata:
-            data['searchterm_pins'] = [
-                t.lower() for t in sdata.searchterm_pins or [] if t]
+            pins = sdata.searchterm_pins or []
```
bduncan137, line 57: getattr to make it safe if property is ever not there?
`pins = getattr(sdata, 'searchterm_pins', [])`

bduncan137 (approved):
there's one possible change in a comment, but otherwise, lgtm

PR #536, opened 2023-01-27
Dh/google analytics latest

DavidHenne:


castle/cms/cron/_social_counts.py
```diff
@@ -213,7 +213,6 @@ def get_social_counts(site, obj, site_url, count_types, count=0):
 
     transaction.commit()
 
-    index_in_es(obj)
```
zombified, line 216: This method was moved to: https://github.com/castlecms/castle.cms/blob/5c859a61bcc7550e51e3a0f6f0e9511f2581dfaa/castle/cms/indexing/hps.py#L159

However, this cron script probably needs additional changes to ensure that it uses WildcardHPS, or maybe a plone interface of some sort, to reindex an object in the catalog, because I don't think this is going to index the object in a custom index for this script, right? it'll just reindex the object in the normal catalog?

either way, update the location of the method here, but don't remove it outright for now. We'll create a different task to fix this.

castle/cms/cron/_twitter_monitor.py
```diff
@@ -82,7 +81,6 @@ def parse_line(site, public_url, line):
                 record_tweet_stats(site, ob, tweet)
     if modified:
         transaction.commit()
-        index_in_es()
```
zombified, line 85: my comment for the same change in the _social_counts.py file applies here -- this should be reverted for now, and the import location updated.

castle/cms/files/aws.py
```diff
@@ -1,13 +1,12 @@
...
 # noqa because these need to precede other imports
 from future.standard_library import install_aliases
-install_aliases()  # noqa
 
 import logging
 from datetime import datetime
 import StringIO
 from time import time
-from urllib.parse import urlsplit, urlunsplit, quote, quote_plus
+import six.moves.urllib as urllib
```
zombified, line 9: why the changes to the lib and imports here in this file? how does this relate to the GA4 update?

castle/cms/files/aws.py
```diff
@@ -20,6 +19,7 @@
 from zope.component import getUtility
 
 
+install_aliases()  # noqa
```
zombified, line 22: this should go back to its original position so that future.standard_library aliases can be correctly applied before other imports are declared

castle/cms/indexing/hps.py
```diff
@@ -156,13 +156,6 @@ def gen_audit_query(
     return query
 
 
-def index_in_es(obj):
```
zombified, line 159: see my comments regarding this change on the _social_counts.py and _twitter_monitor.py files

castle/cms/profiles/default/metadata.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <metadata>
-  <version>3004</version>
+  <version>3005</version>
```
zombified, line 3: this version may have to be updated based on what's currently in master

castle/cms/profiles/default/registry/controlpanel.xml
```diff
@@ -118,4 +118,40 @@
     </value>
   </record>
 
+  <record name="castle.es_index"
```
zombified, line 121: why are the castle.es_index and castle.es_index_enabled being re-added to the registry ?

castle/cms/services/google/analytics.py
```diff
@@ -9,20 +9,27 @@ def get_ga_service(ga_scope=['https://www.googleapis.com/auth/analytics.readonly
...
+    if not api_key or not api_email or (not ga_id and not ua_id):
         return
-
-    # Authenticate and construct service.
-    return get_service(
-        'analytics', 'v3', ga_scope, b64decode_file(api_key)[1], api_email)
+    if ga_id:
+        return get_service(
+            'analytics', 'v4', ga_scope, b64decode_file(api_key)[1], api_email)
+    elif ua_id:
```
zombified, line 19: the castle.universal_analytics_id registry config item is what is backwards compatible with v3 analytics? the castle.universal_analytics_id is a new field, correct? where was that ID value held before? in the castle.google_analytics_id field? is part of the upgrade step migrating the GA3 compatible key in google_analytics_id field to the universal_analytics_id field?

if not, we should make sure to migrate that value so that after this update is applied, sites will still function correctly without having to go into each site configuration to re-configure the GA3 id

castle/cms/tests/test_archival.py
```diff
@@ -23,6 +22,9 @@
 from plone.app.testing import setRoles
 
 
+install_aliases()  # noqa
```
zombified, line 25: this should be left in it's previous position under the import directly so that it can correctly override imports with aliases from future.standard_library

castle/cms/upgrades.zcml
```diff
@@ -65,4 +65,19 @@
     handler=".upgrades.upgrade_3004"
     profile="castle.cms:default"
     />
+  <genericsetup:registerProfile
+    name="3005"
```
zombified, line 69: this version should be updated because of newer commits to master

castle/cms/upgrades/__init__.py
```diff
@@ -36,3 +36,4 @@ def upgrade(context, logger=None):
 upgrade_3001 = default_upgrade_factory('3001')
 upgrade_3003 = default_upgrade_factory('3003')
 upgrade_3004 = default_upgrade_factory('3004')
+upgrade_3005 = default_upgrade_factory('3005')
```
zombified, line 39: this version should be updated because of newer commits to master

zombified (changes requested):

zombified (approved):

PR #537, opened 2023-01-31
GA4 updates for 2.x branch

DavidHenne:


zombified (changes requested):
version number for the profile will likely need to be updated because of other updates that got merged to the 2.x branch

and this PR has a similar question -- if the universal_analytics_id is a new field, and is meant to maintain backwards compat with GA3, but the original GA3 codes were derived from the google_analytics_id field, then is there a migration step that transfers the code from the google_analytics_id field to the universal_analytics_id field? and if not, then we should add that so that we don't have to go into each site and update the analytics settings to re-add the GA3 compat code

PR #553, opened 2023-04-17
ks/md5

katiewiessfelt:


castle/cms/_scripts/templates/watch-run.py
```diff
@@ -29,12 +29,16 @@ def check(directory):
             if check(filepath):
                 changed = True
         else:
-            hashed = md5(filepath)
-            current_hash = times.get(filepath, None)
+            fi = open('/Users/katieschramm/dev/git/FBI/fbigov-dev/sha256_check', 'a')
```
zombified, line 32: hardcoded path?

castle/cms/_scripts/templates/watch-run.py
```diff
@@ -29,12 +29,16 @@ def check(directory):
...
+            path = 'castle/cms/_scripts/templates/watch-run.py check'
```
zombified, line 33: what's this doing?

castle/cms/_scripts/templates/watch-run.py
```diff
@@ -29,12 +29,16 @@ def check(directory):
...
+            fi.write(path + '\n')
+            fi.close()
+            encrypted = sha256(filepath)
```
zombified, line 36: why call it encrypted instead of hashed?

castle/cms/browser/content/__init__.py
```diff
@@ -186,12 +186,19 @@ def chunk_upload(self):
...
                 raise Exception('No tmp upload file found')
         fi = open(info['tmp_file'], mode)
-
-        while True:
-            data = self.request.form['file'].read(2 << 16)
-            if not data:
-                break
-            fi.write(data)
+        type_, location = self.get_type_and_location(self.request.form)
+        if type_ == 'Image':
```
zombified, line 190: maybe instead of duplicating the code between if/else blocks, this could be condensed down to a single block like the removed code, except with the form[] key value being dynamic

eg

```
formtype = 'file'
if type_ == 'Image':
    formtype = 'image'

# ...
data = self.request.form[formtype].read(2 << 16)
# ..
```

castle/cms/browser/dashboard.py
```diff
@@ -150,7 +150,7 @@ def get_modifier(self, item):
         try:
             modifier = item.last_modified_by
             try:
-                # to check if it is hashable and correct type
+                # to check if it is encryptable and correct type
```
zombified, line 153: why encryptable?

castle/cms/browser/security/chat.py
```diff
@@ -33,7 +28,11 @@ def __call__(self):
             for key in keyring:
                 if key is None:
                     continue
-                value = hmac.new(key, user + salt, sha).hexdigest()
+                fi = open('/Users/katieschramm/dev/git/FBI/fbigov-dev/sha256_check', 'a')
```
zombified, line 31: seems like more debug stuff with hard coded paths?

castle/cms/commands.py
```diff
@@ -195,54 +195,34 @@ def __call__(self, filepath):
...
+    To get sha256 encryption of files on the filesystem so
     large files do not need to be loaded into
     memory to be checked
     """
     if os.name == 'nt':
-        bin_name = 'md5.exe'
+        bin_name = 'shasum.exe'
     else:
-        bin_name = 'md5'
+        bin_name = 'shasum'
```
zombified, line 207: isn't it "sha256sum" ? using just `shasum` with no parameters might default to sha1

zombified (changes requested):
in general, why the switch to 'encrypt' or 'encrypted' from 'hash' or 'hashed' everywhere? sha256 is still a hash, and we aren't cryptographically encrypting anything as far as I can tell?

castle/cms/media/video.py
```diff
@@ -35,11 +35,11 @@ def process(context):
 
     # by default, assume all non-mp4 videos need to be converted
     # but in reality, all videos need converting, even mp4.
-    # md5 is only what makes this possible
```
zombified, line 38: seems like the specific hash wouldn't matter here, just that hashing is used. but this comment could be read like md5 specifically is what makes this particular thing possible -- am I missing anything about the specific use of md5 with the logic here? does it interact with the video conversion in any way?

castle/cms/static/plone-compiled.js
```diff
@@ -30165,5 +30167,5 @@ $(document).ready(function(){
 
 });
 
-define("/Users/brian.duncan/castle-instances/pages-to-folders/castle/cms/static/plone.js", function(){});
+define("/Users/katiewiessfelt/dev/git/castle/fbigov-dev/src/castle.cms/castle/cms/static/plone.js", function(){});
```
zombified, line 30170: I wonder if there is a way to remove the specific paths for these from the compiled file?

castle/cms/tasks/files.py
```diff
@@ -87,10 +87,10 @@ def youtube_video_edited(obj):
 
         # we can only edit if file has NOT changed
         # otherwise, it will be reuploaded and original deleted
-        if md5 is not None:
+        if sha256 is not None:
```
zombified, line 90: will this change mean that ALL files would need to be reprocessed?

we may need to keep md5 here (with the 'usedforsecurity=False' param) or figure out a migration path that doesn't involve reprocessing all files?

zombified (changes requested):

castle/cms/archival.py
```diff
@@ -323,7 +323,7 @@ def __init__(self, site, UrlOpener=SubrequestUrlOpener):
         except AttributeError:
             self.archives = self.site._archives = OOBTree()
             self.path_to_uid = self.site._archives_path_to_uid = OOBTree()
-        self.resources = {}  # path -> md5 dict
+        self.resources = {}  # path -> sha-256 dict
```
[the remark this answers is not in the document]
katiewiessfelt: I created a script (`md5_to_sha256.py`) to update the aws resources that should fix this issue

castle/cms/browser/security/chat.py
```diff
@@ -33,7 +28,7 @@ def __call__(self):
             for key in keyring:
                 if key is None:
                     continue
-                value = hmac.new(key, user + salt, sha).hexdigest()
+                value = hmac.new(key, user + salt, sha256).hexdigest()
```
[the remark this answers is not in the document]
katiewiessfelt: the token is being generated by rocket chat. I have changed the code back to how it originally was

castle/cms/media/video.py
katiewiessfelt: I removed the 2 bottom lines of the comments. After looking over the code I decided they don't really make sense since it say that all videos need converting, even mp4s, but it explicitly does not convert mp4s and I can find nowhere in the code where it handles the conversion of mp4s after the `process` method

castle/cms/tasks/files.py
katiewiessfelt: I reimplement md5 and added a `useforsecurity` attribute

castle/cms/static/plone-compiled.js
katiewiessfelt: I couldn't find an easy way to do this, but I can keep looking into it.

PR #557, opened 2023-05-21
Dh/ga4 upgrade

DavidHenne:


castle/cms/browser/content/analytics.py
```diff
@@ -38,29 +44,41 @@ def ga_api_call(self, paths):
...
+            form_type = self.request.form['type']
+            if '-ua' not in form_type:
+                ga_id = registry.get('castle.google_analytics_id', None)
+                service_key = registry.get('castle.google_api_service_key_file', None)
+
+                if ga_id:
+                    form = self.request.form
+                    # Set GOOGLE_ANALYTICS_IS_DEV env variable to true to use mock return data
+                    if os.environ.get("GOOGLE_ANALYTICS_IS_DEV", False):
+                        return get_mock_ga4_data(paths, form)
```
zombified, line 56: a dev toggle like this should be moved into the get_ga4_data method itself so the code here doesn't have to concern itself with whether or not it is in a dev mode.

castle/cms/services/google/__init__.py
```diff
@@ -45,3 +54,101 @@ def get_service(api_name, api_version, scope, key=None,
...
+        service_key = service_key.split(':')[-1]
+        service_key = base64.b64decode(service_key)
+        environ['GOOGLE_ANALYTICS_SERVICE_KEY'] = service_key
+
+    return ga4_run_script(environ)
+
+
+def ga4_run_script(environ):
+    output = None
+    command = ['/usr/bin/python3', 'scripts/google/google-api.py']
```
zombified, line 120: the paths here used for the command should each be configurable in an env var

zombified (changes requested):
in addition to my other comments, we'll also want a PR for the 2.x branch

PR #562, opened 2023-06-22
Ks/request-access

katieschramm:
allows users to request access to the website


castle/cms/browser/security/templates/request-form.pt
```diff
@@ -0,0 +1,72 @@
...
+      <label for="phone">Phone Number:</label><br/>
+      <input type="text" id="phone" name="phone" required><br/>
+      <div class="text-center">
+        <input type="submit" id="submit-request" value="Submit" required><br/>
+      </div>
+      
+    </form>
+    <script>
+      document.getElementById('access-form').addEventListener('submit', (event) => {
+        debugger;
```
zombified, line 45: debugger statement probably shouldn't be here

castle/cms/shield.py
```diff
@@ -67,6 +70,9 @@ def protect(req, recheck=False):
                     anonymous = True
         else:
             anonymous = api.user.is_anonymous()
-
         if anonymous:
-            raise Redirect(login_url)
+            access_setting = registry.get('plone.request_access', False)
+            if access_setting:
+                pass
```
zombified, line 76: IF should protect, IF user is anonymous, IF the access_setting is TRUE, then let the user see whatever page they are requesting?

Seems like that is the logical flow of this code change, unless I'm misreading it -- I don't think we want anon users seeing anything if "protect" is enabled. If they are anonymous and getting to the registry page, then the ISecureLoginAllowedView check above should be what allows them to see anything, right?

zombified (changes requested):

castle/cms/shield.py
katieschramm: but don't we want anonymous users to get through? because won't users be anonymous until they are granted access and the page they are allowed to see is the page where they request access?

zombified (changes requested):
you will need an upgrade step to add the request_access and form_fields in the ISecuritySchema to an existing site

castle/cms/authentication.py
```diff
@@ -89,6 +89,19 @@ def two_factor_enabled(self):
...
+    def request_access(self):
+        if not self.is_zope_root and self.registry:
+            return self.registry.get('plone.request_access_enabled', False)
+        return False
+
+    @property
+    def request_form_url(self):
+        url = '/@@request-form'
+        if not self.is_zope_root and self.registry:
+            url = self.registry.get('plone.request_form_url', '/@@request-form')
```
zombified, line 102: is this plone.request_form_url anymore, or should this be plone.request_access_form_path now?

castle/cms/browser/security/templates/secure-login.pt
```diff
@@ -38,7 +38,13 @@
           </div>
         </div>
       </div>
-
+      <h5 tal:define="url view/request_form_url;"
+           class="text-center"
+           style="color: green">
+        <div tal:condition="python: request.get('HTTP_REFERER').endswith(url)">
```
zombified, line 44: what is this trying to do? Seems like putting this condition here is basically saying "if the request indicates it's come from a url that ends with the full url of the request form, then lets assume the form was submitted" which breaks down to "show a success message after a successful send" -- couldn't this come from maybe a query parameter that indicates success/failure? IE a parameter that is 'submitted' and has a value of 't' if successful? or something along those lines?

zombified (changes requested):

zombified (approved):

PR #573, opened 2023-11-09
Bd/2.x add email templates

bduncan137:
- Update control panel icons, providing one for every castle configlet
- Update dexterity types so every one has an icon and a short desription
- Add EmailTemplate dexterity type for repeated emails
- Fix data-tiles that don't exist from rendering and spamming output
  (for dexterity type definitions in control panel in /@@dexterity-types/<DexterityTypeHere>)
- Allow resuming importjson script


bduncan137 (commented):
I approve this pr

zombified (approved):

PR #581, opened 2023-12-20
Add auto-publish/auto-retract cron script

miknevinas:
This script auto-publishes and retracts content based on specified dates in the 'Dates' tab.

It is assumed that this cron job will run every 30 minutes, as dates can be specified in 30-minute intervals.

The 'admin' user needed to be set to query private documents locally, this can probably be removed.

### DIFF
--- a/castle/cms/cron/_auto_publish_retract.py
+++ b/castle/cms/cron/_auto_publish_retract.py
@@ -1,0 +1,78 @@
+
+from DateTime import DateTime
+from plone import api
+from Products.CMFPlone.interfaces.siteroot import IPloneSiteRoot
+from tendo import singleton
+from zope.component.hooks import setSite
+
+import logging
+import transaction
+
+
+logger = logging.getLogger('castle.cms')
+
+
+def run_query(query_params):
+
+    catalog = api.portal.get_tool('portal_catalog')
+    content = catalog(**query_params)
+    
+    to_state = 'published' if query_params.get('effective') else 'private'
+    for brain in content:
+        obj = brain.getObject()
+        api.content.transition(
+            obj=obj, 
+            to_state=to_state
+        )
+        obj.setModificationDate()
+        obj.reindexObject(idxs=['modified'])
+    transaction.commit()
+
+
+def set_queries(site):
+    setSite(site)
+    
+    # XXX: Needed to run locally, probably not necessary for live environment
+    # from AccessControl.SecurityManagement import newSecurityManager
+    # admin = api.user.get(username='admin')
+    # newSecurityManager(None, admin)
+
+    # publish pending content with an effective date within 30 minutes of cron run
+    start = DateTime() - (30.0 / (24 * 60))
+    end = DateTime()
+    publish_query = {
+        'effective': {
+            'query': (start, end),
+            'range': 'min:max'
+        },
+        'review_state': ['pending', 'private'],
+        'sort_on': 'modified',
+        'sort_order': 'reverse'
+    }
+    run_query(publish_query)
+
+    # retract published content with an expiration date older than cron run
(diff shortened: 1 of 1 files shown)


castle/cms/cron/_auto_publish_retract.py
```diff
@@ -0,0 +1,74 @@
...
+    start = DateTime() - (30.0 / (24 * 60))
+    end = DateTime()
+    publish_query = {
+        'effective': {
+            'query': (start, end),
+            'range': 'min:max'
+        },
+        'review_state': ['pending', 'private']
+    }
+    run_query(publish_query)
```
zombified, line 45: I wonder if this should be named to indicate that it has side effects, and does just get a result -- eg "run_transitions" or something along those lines?

castle/cms/cron/_auto_publish_retract.py
```diff
@@ -0,0 +1,74 @@
...
+        obj.setModificationDate()
+        obj.reindexObject(idxs=['modified'])
+    transaction.commit()
+
+
+def set_queries(site):
+    setSite(site)
+
+    # publish pending content with an effective date within 30 minutes of cron run
+    start = DateTime() - (30.0 / (24 * 60))
```
zombified, line 36: I'm not sure I follow the `30.0 / (24 * 60)` bit -- what value does represent?

I apologize if this is obvious, I'm probably just be dense. I would expect "30 minutes ago" to be something more like `DateTime() - (30.0 * 60)`

or maybe a python standard library take on it (eschewing the zope DateTime), like:

```
datetime.timestamp(datetime.now() - timedelta(minutes=30))
```

unless there's some hidden timezone thing that happens with the zope DateTime (I don't know off hand)

zombified (changes requested):
I think we also want a method like https://github.com/castlecms/castle.cms/blob/master/castle/cms/cron/__init__.py#L119-L120 in the castle/cms/cront/__init__.py file

and a reference to that method like https://github.com/castlecms/castle.cms/blob/master/setup.py#L135 in the setup.py file

so we can have a bin script generated for it

castle/cms/cron/_auto_publish_retract.py
miknevinas: Noted, changed for clarity.

castle/cms/cron/_auto_publish_retract.py
miknevinas: No worries, it's definitely strange. This is the syntax I found to subtract 30 minutes from a Zope DateTime object. The math evaluates to '0.02083333', which is the exact value needed to subtract 30 minutes from the object. I have no idea how this resembles 30 minutes, but it makes sense to Zope. Here are the start/end dates when running:

`
2024/01/08 07:37:49.145811 HST
2024/01/08 08:07:49.145848 HST
`

I could replace the math with a hardcoded float, though I think it would make it more confusing.

I could also convert it to a standard python datetime object for legibility and then use timedelta, though I think it would need to be converted back to Zope for the query.

As for timezone, I found DateTime() utilized similarly in other cron scripts without any timezone adjustments, so I assume Zope does something under the hood for that.

zombified: is there a link to the discussion where you found this?

miknevinas: I couldn't find the discussion, but after digging into the DateTime module I found a method that converts it to epoch time, so now it just subtracts 30 minutes from epoch and re-converts it into DateTime, much more sensible logic. My fault for not finding that method earlier.

zombified: awesome, thank you!

zombified (approved):

PR #584, opened 2024-01-30
Sharing permissions fix

miknevinas:
The 'Plone/image-repository/@@sharing' view on FBIGOV would throw a 'maximum recursion depth exceeded' error when attempting to grant users sharing permissions. The snag occurred in the cPickle library. The simplest solution was to temporarily increase the python recursion limit so the operation could be completed.


miknevinas:
Closed due to changes here being added to a separate PR

